PR #36, opened 2023-01-08
Add gc support

lenisko:
I'll add GDS soon.

### DIFF
--- a/routes.go
+++ b/routes.go
@@ -63,6 +63,8 @@ func Raw(c *gin.Context) {
 	// than I would like
 
 	pogodroidHeader := r.Header.Get("origin")
+	userAgent := r.Header.Get("User-Agent")
+
 	if pogodroidHeader != "" {
 		var raw []map[string]interface{}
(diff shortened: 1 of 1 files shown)


routes.go
```diff
@@ -112,19 +114,40 @@ func Raw(c *gin.Context) {
...
 				entry := v.(map[string]interface{})
-				protoData = append(protoData, InboundRawData{
-					Base64Data: entry["payload"].(string),
-					Method:     int(entry["type"].(float64)),
-					HaveAr: func() *bool {
-						if v := entry["have_ar"]; v != nil {
-							res, ok := v.(bool)
-							if ok {
+				// Atlas, GC and GDS support
+				if userAgent[:13] == "Pokemod Atlas" || userAgent[:10] == "PokmonGO/0" {
```
jfberry, line 118: I think this will fail if user agent is shorter or missing than 13 chars. Should also check for length for safety.

lenisko: Thanks! I'll add check later

lenisko:
Fixed 👍 merge anytime

PR #39, opened 2023-01-18
fix db clean rows & time

lenisko:
I'll test it soon


lenisko:
Works well, ready to merge @jfberry

PR #41, opened 2023-01-19
Replace player ID by auto incremented int

na-ji:
The goal of this PR is to make the player table compatible with _another tool_.

### DIFF
--- a/decoder/player.go
+++ b/decoder/player.go
@@ -15,8 +15,9 @@ import (
 )
 
 type Player struct {
-	Id                 null.String `db:"id"`
+	// Name is the primary key
 	Name               string      `db:"name"`
+	FriendshipId       null.String `db:"friendship_id"`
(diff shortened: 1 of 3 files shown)


jfberry:
The problem with this is that if another tool is holding this data, and sending raws to golbat the ids will not be aligned - since it is an auto-increment field

na-ji:
> The problem with this is that if another tool is holding this data, and sending raws to golbat the ids will not be aligned - since it is an auto-increment field

Golbat is using the username to find and update player data. So how is it an issue?

jfberry:
Is this confirmed as the way you want to go? Or holding out for recent raid changes?

na-ji:
No it's not, hence why I made the PR a draft. In the end, I will go this way: 
https://discord.com/channels/973641026117972038/1023686446671999046/1065895293939372104
But for that, I need to be able to parse the request

PR #44, opened 2023-01-28
Store S2 cells

Fabio1988:
@unseenmagik thanks for initial contribution :) 

Store GMO map cells

Update `updated` timestamp of pokestop, gym, s2cell regularly in a certain time windows (every 15 minutes)

Clear Old Forts after 30 min unseen

### DIFF
--- a/main.go
+++ b/main.go
@@ -423,12 +423,6 @@ func decodeGMO(ctx context.Context, sDec []byte, username string) string {
 			case pogo.FortType_GYM:
 				gymIdsPerCell[mapCell.S2CellId] = append(gymIdsPerCell[mapCell.S2CellId], fort.FortId)
 			}
-			if fort.FortId == "6a267847d2934d199fa9a50ee54e270b.16" || fort.FortId == "880ac223bc0041dead1bba87fe7f76cd.16" {
-				log.Infof("Found stop: %s", fort.FortId)
-			}
-		}
-		if mapCell.S2CellId == 5146962338993143808 || mapCell.S2CellId == 5146778065568792576 {
-			log.Infof("Map Cell stops count: %v", len(stopIdsPerCell[mapCell.S2CellId]))
 		}
 		newClientMapCellS2CellIds = append(newClientMapCellS2CellIds, mapCell.S2CellId)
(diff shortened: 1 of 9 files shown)


main.go
```diff
@@ -413,7 +416,14 @@ func decodeGMO(ctx context.Context, sDec []byte, username string) string {
 			if fort.ActivePokemon != nil {
 				newMapPokemon = append(newMapPokemon, decoder.RawMapPokemonData{Cell: mapCell.S2CellId, Data: fort.ActivePokemon})
 			}
+			// collect fort information of cell
+			if fort.FortType == pogo.FortType_CHECKPOINT {
```
Fabio1988, line 420: @jfberry correct me if that is the wrong approach in Golang to compare ENUM Literals :)
is there an exhaustive way to compare?

In Kotlin /Java you could do something like that:
```kotlin
when (fort.FortType) {
  pogo.FortType_CHECKPOINT -> ....
  pogo.FortType_GYM -> ...
}
```

Fabio1988:
I would appreciate your review guys :) not tested yet, need to setup a non-production environment first :+1:

main.go
jfberry: switch statement is what you are looking for
https://gobyexample.com/switch

main.go
```diff
@@ -413,7 +416,15 @@ func decodeGMO(ctx context.Context, sDec []byte, username string) string {
 			if fort.ActivePokemon != nil {
 				newMapPokemon = append(newMapPokemon, decoder.RawMapPokemonData{Cell: mapCell.S2CellId, Data: fort.ActivePokemon})
 			}
+			// collect fort information of cell
+			switch fort.FortType {
+			case pogo.FortType_CHECKPOINT:
+				stopIdsPerCell[mapCell.S2CellId] = append(stopIdsPerCell[mapCell.S2CellId], fort.FortId)
```
Fabio1988, line 422: is it possible to collect that information in that way @jfberry ?

```go
var stopIdsPerCell = make(map[uint64][]string)
```
appending the map value with content, or is it needed to initialize the slices? :)

PR #49, opened 2023-02-05
Handle GetMapForts

lenisko:
Handle `METHOD_GET_MAP_FORTS`

TODO:
- [x] GetMapForts Cache
- [x] Gym Mutex Lock

### DIFF
--- a/decoder/gym.go
+++ b/decoder/gym.go
@@ -513,7 +513,7 @@ func UpdateGymRecordWithGetMapFortsOutProto(db db.DbDetails, mapFort *pogo.GetMa
 	// we missed it in Pokestop & Gym. Lets save it to cache
 	if gym == nil {
 		getMapFortsCache.Set(mapFort.Id, mapFort, ttlcache.DefaultTTL)
-		log.Debugf("Saved getMapFort in cache: %s", gym.Id)
+		log.Debugf("Saved getMapFort in cache: %s", mapFort.Id)
 		return false, ""
 	}
 
--- a/decoder/gym.go
+++ b/decoder/gym.go
@@ -449,6 +449,16 @@ func saveGymRecord(db db.DbDetails, gym *Gym) {
 	createGymWebhooks(oldGym, gym)
 }
 
+func updateGymGetMapFortCache(gym *Gym) {
+	storedGetMapFort := getMapFortsCache.Get(gym.Id)
+	if storedGetMapFort != nil {
+		getMapFort := storedGetMapFort.Value()
+		getMapFortsCache.Delete(gym.Id)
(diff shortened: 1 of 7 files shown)


decoder/gym.go
```diff
@@ -446,12 +473,18 @@ func UpdateGymRecordWithFortDetailsOutProto(db db.DbDetails, fort *pogo.FortDeta
 		gym = &Gym{}
 	}
 	gym.updateGymFromFortProto(fort)
+
+	updateGymGetMapFortCache(gym)
```
jfberry, line 477: FortDetailsOut contains name anyway doesn't it? I guess don't want to overwrite those (would need verifying that they are actually in there not just possibly in there, I can't remember)

decoder/gym.go
```diff
@@ -461,6 +494,30 @@ func UpdateGymRecordWithGymInfoProto(db db.DbDetails, gymInfo *pogo.GymGetInfoOu
 		gym = &Gym{}
 	}
 	gym.updateGymFromGymInfoOutProto(gymInfo)
+
+	updateGymGetMapFortCache(gym)
```
jfberry, line 498: As above, does this contain the name anyway?

main.go
```diff
@@ -340,6 +343,40 @@ func decodeFortDetails(ctx context.Context, sDec []byte) string {
...
+
+	var outputString string
+	processedForts := 0
+
+	for _, fort := range decodedMapForts.Fort {
+		// when we miss, we check the gym, if again, we save it in cache for 5 minutes (in gym part)
+		status, output := decoder.UpdatePokestopRecordWithGetMapFortsOutProto(ctx, dbDetails, fort)
+		if !status {
+			status, output = decoder.UpdateGymRecordWithGetMapFortsOutProto(dbDetails, fort)
+		}
```
jfberry, line 367: I think it should add to the cache here not as magic inside the UpgradeGymRecord.

decoder/gym.go
lenisko: They were there in all cases I tested, but can do some length check if that's what you mean. You can rename fort I guess so that still would make sense.

jfberry: What I mean is that you have just received a raw which contains the name I am not sure you should overwrite it with the cached value

lenisko: Hmm.. okay make sense 👍

main.go
jfberry: getMapFortsCache would be available to whole of the module surely?
It's magic because the two methods do something different. One just tries an update on a pokestop, the other tries the update on the gym, but if it doesn't work it invisibly adds it to the cache

jfberry (commented):
Lgtm

PR #50, opened 2023-02-07
Webhook areas

Fabio1988:
Still WIP, needs testing, needs comparing of performance etc.

Create PR to see diff better


webhooks/collector.go
```diff
@@ -109,3 +176,40 @@ func collectHooks() []WebhookQueue {
 
 	return destinations
 }
+
+func doAreasMatch(messageAreas, hookAreas []string) bool {
+	for _, area1 := range hookAreas {
+		hookArea, hookFence := splitIntoAreaAndFenceName(area1)
```
jfberry, line 182: This should have been split up on load of the config; this is being split every single webhook

webhooks/collector.go
```diff
@@ -70,27 +73,91 @@ func collectHooks() []WebhookQueue {
...
+				for _, message := range currentCollection[GymDetails].Messages {
+					if doAreasMatch(message.Areas, hook.Areas) {
+						totalCollection = append(totalCollection, message)
+					}
+				}
+			}
 		}
 		if hook.Types == nil || slices.Contains(hook.Types, "raid") {
-			totalCollection = append(totalCollection, currentCollection[Raid].Messages...)
+			if hook.Areas == nil || len(hook.Areas) == 0 {
```
jfberry, line 91: len(nil) == 0 in go so you can just check len(hook.Areas) == 0

webhooks/collector.go
```diff
@@ -70,27 +73,91 @@ func collectHooks() []WebhookQueue {
 
 	for _, hook := range config.Config.Webhooks {
 
+		//TODO: check for area matching for each hook message
+
 		var totalCollection []WebhookMessage
 		if hook.Types == nil || slices.Contains(hook.Types, "gym") {
-			totalCollection = append(totalCollection, currentCollection[GymDetails].Messages...)
+			if hook.Areas == nil || len(hook.Areas) == 0 {
```
jfberry, line 80: len(nil array) in go is valid = 0
so you can just do `if len(hook.Areas) == 0` I would think

webhooks/collector.go
```diff
@@ -49,11 +51,12 @@ func SetMaps() {
 	webhookCollections[Weather] = &WebhookList{}
 }
 
-func AddMessage(webhookType string, message interface{}) {
+func AddMessage(webhookType string, message interface{}, areas []string) {
```
jfberry, line 54: I would think you would want to pass the areaNames as the expanded structure, not compress into a string only to split this out again

Fabio1988:
This PR needs to find the matching area/fence with every single element (pokemon, pokestop, invasion, gym, weather,....). this might be expensive over time. If someone is not using webhook areas overall this might be some overhead we could reduce with a separate config `enableWebhookAreas: false` ?!

webhooks/collector.go
Fabio1988: yeah thought that too... but did not really find a solution for that. geography.go and stats.go are both in decoder module.

also stats.go relates to pokemon struct etc... maybe you have some time to figure that out? as I told you keeping the struct would lead into a cyclic dependey instead :(

jfberry: Move areaName and the match definition into the geo module
the loader can stay in geography as that is a specific use I guess (or there could be a ReadGeoJson(filename) which returns the structure which geography then becomes much simpler

Fabio1988: done, thanks for the hint

webhooks/collector.go
Fabio1988: done

config/reader.go
```diff
@@ -43,4 +45,20 @@ func ReadConfig() {
 	if err != nil {
 		panic(err)
 	}
+	// translate webhook areas to array of geo.AreaName struct
+	for _, hook := range Config.Webhooks {
```
Fabio1988, line 49: @jfberry it seems that this "workaround" in loading config does not write the AreaNames into the webhook config.

So in webook sending the hook.AreaNames is empty, but the hook.Areas are set

jfberry: That is because hook is a copy, not the original value.
Try
```
for i := 0; i < len(Config.Webhooks); i++ {
  hook := &Config.Webhooks[i]
  hook.AreaNames = splitIntoAreaAndFenceName(hook.Areas)
}
```

Fabio1988: thank you that worked :)

Fabio1988:
webhook filtering for areas is working, thanks to @dkmur for testing :)

can someone confirm pvp webhooks are still working? or quest webhooks? :)

dkmur:
> webhook filtering for areas is working, thanks to @dkmur for testing :)
> 
> can someone confirm pvp webhooks are still working? or quest webhooks? :)

checked both webhooks received poracle side and look fine.
PVP messsages are posted to discord, quest are not posted to channel nor do I track anything so dunno.
No user complaints so far and generally such things are reported quite quickly.

PR #52, opened 2023-02-20
Decode social protos depending on request

na-ji:
Hello,

I've made a little change to the way Golbat is saving social protos. From now on, it will decode the protos according to the request. The request contains the social action ID, or the friend code, which allows to correctly save the player.


jfberry:
This looks visually fine...

PR #54, opened 2023-02-22
Forts webhook

### LINKED ISSUE #48: Define webhook for cleared/new forts
We should define them somehow :)

Fabio1988:
closes #48 

NEW:
```json
{
  "type": "fort_update",
  "message": {
    "change_type": "new",
    "new": {
      "type": "pokestop",
      "id": "1a68ecfa0a952ef8a5912192527a40ec.16"
      "name": "Brunnen",
      "description": "keine",
      "image_url": "http://lh3.googleusercontent.com/dYgMDyDD2g9V8kTwuSkUvyEL5WNVCWhuZejj_9oGWG1ViCWOm34WOe-9rx1WYRQYjtMAO80YU5cpn5r0aYzUKf7H1Lo",
      "location": {
        "lat": 47.26478,
        "lon": 11.40795
      }
    }
  }
}
```
```json
{
  "type": "fort_update",
  "message": {
    "change_type": "new",
    "new": {
      "type": "pokestop",
      "id": "1a68ecfa0a952ef8a5912192527a40ec.16"
      "name": null,
      "description": null,
      "image_url": null,
      "location": {
        "lat": 47.26478,
        "lon": 11.40795
      }
    }
  }
}
```
REMOVAL:
```json

{
    "type": "fort_update",
    "message": {
      "change_type": "removal",
      "old": {
        "id": "3a68ecfa0a954ef8a5912192527a40eb.16",
        "type": "pokestop",
        "name": "Stegosawwwwrus Column",
        "description": "",
        "image_url": "http://lh3.googleusercontent.com/031cZn9HMC8F6uLdt3czf4ZpB0A3AYjQtARaoWI9uz2lLfQATRdgtjZ81vJ9KwvZ0ubt4tYQSKGs5psTHrZX4dyPp6A",
        "location": {
          "lat": 48.160017,
          "lon": 16.323872
        }
      }
    }
}

```
EDIT:
```json
{
  "type": "fort_update",
  "message": {
    "change_type": "edit",
    "edit_types": ["description"],
    "old": {
      "type": "pokestop",
      "id": "1a68ecfa0a952ef8a5912192527a40ec.16"
      "name": "Brunnen",
      "description": "keine",
      "image_url": "http://lh3.googleusercontent.com/dYgMDyDD2g9V8kTwuSkUvyEL5WNVCWhuZejj_9oGWG1ViCWOm34WOe-9rx1WYRQYjtMAO80YU5cpn5r0aYzUKf7H1Lo",
      "location": {
        "lat": 47.26478,
        "lon": 11.40795
      }
    },
    "new": {
      "type": "pokestop",
      "id": "1a68ecfa0a952ef8a5912192527a40ec.16"
      "name": "Brunnen",
      "description": "das ist die neue Beschreibung",
      "image_url": "http://lh3.googleusercontent.com/dYgMDyDD2g9V8kTwuSkUvyEL5WNVCWhuZejj_9oGWG1ViCWOm34WOe-9rx1WYRQYjtMAO80YU5cpn5r0aYzUKf7H1Lo",
      "location": {
        "lat": 47.26478,
        "lon": 11.40795
      }
    }
  }
}
```

### DIFF
--- a/decoder/fort.go
+++ b/decoder/fort.go
@@ -9,7 +9,6 @@ import (
 	"golbat/geo"
 	"golbat/pogo"
 	"golbat/webhooks"
-	"gopkg.in/guregu/null.v4"
 )
 
 type Location struct {
@@ -18,11 +17,11 @@ type Location struct {
 }
 
 type FortWebhook struct {
-	Type        null.String `json:"type"`
-	Name        null.String `json:"name"`
-	Description null.String `json:"description"`
-	ImageUrl    null.String `json:"image_url"`
-	Location    Location    `json:"location"`
+	Type        string   `json:"type"`
+	Name        *string  `json:"name"`
+	Description *string  `json:"description"`
+	ImageUrl    *string  `json:"image_url"`
+	Location    Location `json:"location"`
 }
 
 type FortChange string
@@ -63,10 +62,10 @@ func InitWebHookFortFromGym(gym *Gym) (fort FortWebhook) {
 	if gym == nil {
 		return
 	}
-	fort.Type = null.StringFrom(GYM.String())
-	fort.Name = gym.Name
-	fort.ImageUrl = gym.Url
-	fort.Description = gym.Description
+	fort.Type = GYM.String()
+	fort.Name = gym.Name.Ptr()
+	fort.ImageUrl = gym.Url.Ptr()
+	fort.Description = gym.Description.Ptr()
 	fort.Location = Location{Latitude: gym.Lat, Longitude: gym.Lon}
 	return
 }
@@ -75,10 +74,10 @@ func InitWebHookFortFromPokestop(stop *Pokestop) (fort FortWebhook) {
 	if stop == nil {
 		return
 	}
-	fort.Type = null.StringFrom(POKESTOP.String())
-	fort.Name = stop.Name
-	fort.ImageUrl = stop.Url
-	fort.Description = stop.Description
+	fort.Type = POKESTOP.String()
+	fort.Name = stop.Name.Ptr()
+	fort.ImageUrl = stop.Url.Ptr()
+	fort.Description = stop.Description.Ptr()
 	fort.Location = Location{Latitude: stop.Lat, Longitude: stop.Lon}
 	return
 }
--- a/decoder/fort.go
+++ b/decoder/fort.go
@@ -2,10 +2,13 @@ package decoder
 
 import (
 	"context"
+	"encoding/json"
 	"github.com/jellydator/ttlcache/v3"
 	log "github.com/sirupsen/logrus"
 	"golbat/db"
+	"golbat/geo"
 	"golbat/pogo"
+	"golbat/webhooks"
 )
 
 type FortWebhook struct {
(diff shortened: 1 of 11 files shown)


Fabio1988:
@ccev should there be an hint what exactly changed on EDIT ?

ccev:
just in the edit_types array

Fabio1988:
So for edit, I will add a `edit_type`, should be easy :)

jfberry:
I think removal should have 'old' instead of 'fort'
I think new should have 'new' instead of fort

ie - new values always in new, old values always in old

decoder/gym.go
```diff
@@ -102,7 +103,7 @@ type Gym struct {
 //FROM information_schema.columns
 //WHERE table_schema = 'db_name' AND table_name = 'tbl_name'
 
-func getGymRecord(ctx context.Context, db db.DbDetails, fortId string) (*Gym, error) {
+func GetGymRecord(ctx context.Context, db db.DbDetails, fortId string) (*Gym, error) {
```
jfberry, line 106: why do these now need to be public?

Fabio1988: When a gym/stop is cleared I only have the id, but not gym object in general. 

And I need Gym object to access the must have props for hook... Accessing cache might be too less information, that's why I need to access db row too if not in cache

jfberry: But from where? Looks like its the same module so has access anyway

Fabio1988:
> I think removal should have 'old' instead of 'fort' I think new should have 'new' instead of fort
> 
> ie - new values always in new, old values always in old

Done :)

decoder/gym.go
Fabio1988: 🤔 have to try, I think it was not available in code completion

decoder/main.go
```diff
@@ -411,10 +416,11 @@ func ClearRemovedForts(ctx context.Context, dbDetails db.DbDetails,
 						continue
 					}
 					log.Infof("Cleared old Stop(s) in cell %d: %v", cellId, toClear)
-					//TODO send webhook
+					CreateFortWebhooks(ctx, dbDetails, toClear, POKESTOP, REMOVAL)
```
Fabio1988, line 419: @jfberry I don't think it's accessible from here (getPokestopRecord)... I can try again tomorrow

jfberry: it's in decoder as well

Fabio1988: Makes sense, maybe eyes were to tired haha

ccev:
- maybe re-think using `fort` as the wbehook type. maybe a bit too ambigious? `fort_change` might be better
- if the fort is missing name/desc/image it should be null. not sure if that's the case from your examples
- imo the contents edit_types should be consistent with the fort object field names. so instead of lat/lon, a `location` field might be better:
```
{
    "edit_types": ["name", "location"],
    "old": {
        "name": "...",
        "location": {
            "lat": 0.0,
            "lon": 0.0
        }
       ...
    },
    ...
}
```

Fabio1988:
> * maybe re-think using `fort` as the wbehook type. maybe a bit too ambigious? `fort_change` might be better
> * if the fort is missing name/desc/image it should be null. not sure if that's the case from your examples
> * imo the contents edit_types should be consistent with the fort object field names. so instead of lat/lon, a `location` field might be better:
> 
> ```
> {
>     "edit_types": ["name", "location"],
>     "old": {
>         "name": "...",
>         "location": {
>             "lat": 0.0,
>             "lon": 0.0
>         }
>        ...
>     },
>     ...
> }
> ```

Good points :)

* first one easy to change, it's probably not a `fort_change`, it's a `fort_update`?! :)
* missing values as null have to be done, is not done yet
* Already changed location to your suggestion, forgot to update examples, thank you :)

ccev:
`fort_update` sounds good to me

decoder/fort.go
```diff
@@ -2,12 +2,188 @@ package decoder
...
+	Latitude  float64 `json:"lat"`
+	Longitude float64 `json:"lon"`
+}
+
+type FortWebhook struct {
+	Type        null.String `json:"type"`
+	Name        null.String `json:"name"`
+	Description null.String `json:"description"`
+	ImageUrl    null.String `json:"image_url"`
+	Location    Location    `json:"location"`
```
Fabio1988, line 25: @jfberry how to make that struct nullable? :confused:

decoder/fort.go
```diff
@@ -2,12 +2,187 @@ package decoder
...
+		if old.Description != new.Description {
+			editTypes = append(editTypes, "description")
+		}
+		if old.ImageUrl != new.ImageUrl {
+			editTypes = append(editTypes, "image_url")
+		}
+		if old.Location != new.Location {
+			editTypes = append(editTypes, "location")
+		}
+		hook := map[string]interface{}{
```
Fabio1988, line 164:
```golang
hook := map[string]interface{}{
			"change_type": change.String(),
			"edit_types":  editTypes,
			"old": old,
			"new": new,
}
```

is it possible to use it like that? or do we need the json.RawMessage?! :)

jfberry: This works, except old and new would always be present since there is no ability to do a 'omitempty'
If you did it as a structure you could tag old new and edit_types with omitempty
Or you can create a map as you have and just add old and new in if non nil
Except if you want old and new to always be present (ie with a null value) then this work work I would guess

Fabio1988: In that case I already know that there are differences and old and new are never empty :)

jfberry (commented):
lgtm

jfberry:
dammit I now have to write some poracle code :-(

PR #55, opened 2023-02-24
quest status api

TurtIeSocks:
- adds sql function to get list of quests in a geofence
- api endpoint to call the function

### DIFF
--- a/db/pokestop.go
+++ b/db/pokestop.go
@@ -19,9 +19,9 @@ type FortId struct {
 }
 
 type QuestStatus struct {
-	Quests     uint32 `db:"quests"`
-	AltQuests  uint32 `db:"alt_quests"`
-	TotalStops uint32 `db:"total"`
+	Quests     uint32 `db:"quests" json:"quests"`
+	AltQuests  uint32 `db:"alt_quests" json:"alt_quests"`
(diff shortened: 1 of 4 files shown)


jfberry:
This looks like it's the right idea here - I notice that the modules have been re-ordered in the files - is that goland that is doing that or some other editor?
The files were formated by `go fmt` I think, but perhaps this is updated in a newer version of go - just trying to understand what is behind this

Fabio1988:
> This looks like it's the right idea here - I notice that the modules have been re-ordered in the files - is that goland that is doing that or some other editor? The files were formated by `go fmt` I think, but perhaps this is updated in a newer version of go - just trying to understand what is behind this

I used latest goland with default settings and my IDE is not reordering them, so might be external formattwer?!

jfberry:
I can see why it has (external dependencies after internal and library). But it’s curious - wouldn’t want formatting to change with every commit

Fabio1988:
We should definitely share our configs for fomatting, it seems that ours does match and Turtles not :)

TurtIeSocks:
Probably a vsc formatting issue then. Got annoyed with goland and haven't gone back yet.

lenisko:
@Fabio1988 or just add workflow and add post-commit 🤔

PR #60, opened 2023-03-10
align ar/noar quest naming

ccev:
drop stupid quest/alt quest naming

### DIFF
--- a/db/pokestop.go
+++ b/db/pokestop.go
@@ -18,8 +18,8 @@ type FortId struct {
 }
 
 type QuestStatus struct {
-	Quests     uint32 `db:"quests" json:"quests"`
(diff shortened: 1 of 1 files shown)


Fabio1988:
imo that makes code really bad to read in future... We would need to change all occurences, also in the database

jfberry:
We should not change the db schema just yet. Let’s just get the external naming right and fix the schema later

jfberry:
If you wanted to make this clearer you could keep the alt naming for the db translation of the structure and output using the new naming in the json

ccev:
i'm not sure where else this used. this is my quest to fix the naming from Turtle's recent Quest Status PR. AR/NoAR naming is used in Dragonite like this too. Imo it's a lot harder to understand Normal/Alt Quest naming

na-ji (approved):

PR #61, opened 2023-03-13
koji support

TurtIeSocks:
- adds loading stats geofence from koji
- adds 2 config values: `koji_url` and `koji_bearer_token`


jfberry:
Will have a look later but nest fence is actually to be loaded from nest table, generator almost in place

jfberry:
1. The LoadGeofence is currently called from a library init() function. This is called before the application startup.
This will need to be moved into user code, and called after config being loaded.
2. The current model here has the system trying to call koji and falling back to local file. If Koji is unavailable, a golbat restart now fails. This is obviously not good behaviour.

Probably the code (after startup) should:
1.  Attempt to download a geofence from koji, and store it somewhere.  Load this fence and store it somewhere.  If Koji is unavailable, then fall-back to the most recent downloaded fence.  There was a .cache folder created for timezone geofence data, perhaps it is still created (or maybe I removed all traces since it is no longer used)
2. If Koji is not configured, then load the geofence from the standard place.

jfberry:
I moved the fence reloading in the nestsdb branch, which will fix the problem with early init.
I was going to finish nests but I realised just how useless they are, so this branch will probably sit for a while - but you could grab the useful bits out to prevent a merge failure later.  Or build on this (it could be merged as is, since nests was experimental and only logged before - this does the same but reads the nest polygon from a newly created nests table)

TurtIeSocks:
I think it's all set now @jfberry. Let me know if you want me to take out some of the bits from the other branch, particularly the migration.

TurtIeSocks:
Thanks @Fabio1988 . Is this good to go now @jfberry ?

jfberry:
I think there was an issue raised with null area names by @lenisko ?

lenisko:
@jfberry more context please ;-)

decoder/geography.go
```diff
@@ -1,27 +1,103 @@
...
+}
 
 func ReadGeofences() error {
+	if kojiUrl != "" {
+		fc, err := GetKojiGeofence(kojiUrl)
+		if err != nil {
+			log.Warnf("KOJI: Unable to get geofence from koji - %s", err)
+			geofence, err := ioutil.ReadFile(kojiCacheFilename)
+			if err != nil {
+				log.Warnf("KOJI: Unable to read cached geofence - %s", err)
```
jfberry, line 72: if koji url drops through here, and there is no cache this currently drops through to current reader.
Is that correct? Or should it start with no fences if that's the case

config.toml.example
```diff
@@ -5,6 +5,9 @@ api_secret = "golbat"   # Golbat secret required on api calls (blank for none)
 stats = false           # Enable creation of pokemon_timing and pokemon_history table (now recommended=false)
 in_memory = false       # Use in-memory storage for pokemon only
 
+# koji_url = "http://{koji_url}/api/v1/geofence/feature-collection/{golbat_project}"
```
jfberry, line 8: Do we need a koji section? Is this likely to be it for settings?

jfberry:
I'm generally happy with this, comments are suggestions/questions rather than blockers

config.toml.example
TurtIeSocks: Added just in case.

decoder/geography.go
TurtIeSocks: Correct. 

No koji url in config:
```
INFO 2023-03-24 11:45:41 GEO: Loaded geofence from geofence.json
```
When koji is available:
```
INFO 2023-03-24 11:39:06 Setting Koji Info http://0.0.0.0:8080/api/v1/geofence/feature-collection/my_project
INFO 2023-03-24 11:39:06 Initialising Ohbem for PVP
2023/03/24 11:39:06 Cache cleaned
2023/03/24 11:39:06 MasterFile Watcher Started
INFO 2023-03-24 11:39:06 KOJI: Response 200 OK
INFO 2023-03-24 11:39:06 KOJI: Loaded geofence from koji, caching
```

When koji is unavailable, and a cached koji json has been saved:
```
WARN 2023-03-24 11:44:24 KOJI: Unable to get geofence from koji - Get "http://0.0.0.0:8080/api/v1/geofence/feature-collection/my_project": dial tcp 0.0.0.0:8080: connect: connection refused
INFO 2023-03-24 11:44:24 KOJI: Loaded cached geofence
```

When koji is unavailable and no cached json is found:
```
WARN 2023-03-24 11:44:39 KOJI: Unable to get geofence from koji - Get "http://0.0.0.0:8080/api/v1/geofence/feature-collection/my_project": dial tcp 0.0.0.0:8080: connect: connection refused
WARN 2023-03-24 11:44:39 KOJI: Unable to read cached geofence - open cache/koji_geofence.json: no such file or directory
INFO 2023-03-24 11:44:39 GEO: Loaded geofence from geofence.json
```

TurtIeSocks: If this is fine with you @jfberry, I'll merge in.

jfberry: Right, I thought if you were using koji config, and koji unavailable, and no cached version it shouldn't load geofence.json (because thats likely a vestible from who knows when?)
If you are happy that this behaviour is as intended I am happy otherwise.

PR #62, opened 2023-03-15
Stats history

unseenmagik:
Ability to adjust the how long historic stats are kept before being removed from the database.

- New config variable `stats_days = 7` # Where the value defined is in days.

If no value is set, or stats_days is omitted from config.toml, default will be 7 days if `stats = true`


jfberry:
lgtm

unseenmagik:
PR implemented

PR #64, opened 2023-03-16
Nest database support

jfberry:
* Create table for nests
* Load nest fences from table

Unfinished


jfberry:
Current status:
nest branch - golbat creates the nest table, and will read nests from it
It collects and categorises the pokemon; however this really needs to be in hour long buckets (perhaps using time.now() % 3600 as the key) dropping off old ones so that we have multi-hours worth of data which expire - as sample sizes are too small currently.
It needs to pick the pokemon with the highest %age; but not if the pokemon is > 50% spawns in that hour [this would indicate spotlight or community day]
then it just needs to write it to the db in a batch rather than logging it

jfberry:
I will not be working on this branch any further

PR #65, opened 2023-03-22
A better Pokemon parser

### LINKED ISSUE #3: On event change, IV is stripped (just like weather) - IV could be preserved if weather boost not changed for this new mons type

Mygod:
New features:
* Better support of Ditto and weather/spawn rerolls. (Fixes #3.)
* Seeing a NearbyPokemon multiple times will update lat/lon to a more accurate guess.
* Chuck's code structure is employed to reduce duplicate code.

Notes:
* alternativeDisplay is not ported since scanning A/B spawns does not appear to be very useful.
* ~~stats_statsReset is not used any more since IV/level are now never reset.~~
* Capture probabilities are not populated any more since they differ by account.
* Compared to Chuck, `IvInactive` is now compressed to 2 bytes instead of 3 fields since golang actually has integer types unlike a certain maniac language.


Mygod:
Known issues:

* CP is not recalculated when inactive IV is activated. Needs to call `calculateCP` in ohbemgo...

Mygod:
What's new:

* `stats_statsReset` is brought back based on how @Fabio1988 thinks it should work. It should also work nicely with Ditto.
* The ultimate Ditto detector is implemented. 8 more cases are supported than RDM/Chuck. Now additional Ditto can be picked up upon rescan after weather rerolls. Performance impact is minimal.
* Ready for testing.

Mygod:
@lenisko Do you think you could make `calculateCP` public?

lenisko:
@Mygod Feel free to open PR. I'll accept once I'll wake up 👍

Mygod:
New Ditto detections to be tested:

* [ ] 00>0P
* [ ] 00>B0
* [ ] 0P>00
* [ ] 0P>B0
* [ ] B0>00
* [ ] B0>0P
* [ ] B0>PP
* [ ] PP>B0

TurtIeSocks (approved):
running in production across multiple maps

Fabio1988:
Awesome work, should we change the log level of that entry to debug?!
I think we still store it in golbat.logs, that's okay, but to reduce console log output?

```Pokemon 6081190677848492040 changed from (209,0,0,3) to (209,1253,0,2)```

Fabio1988 (approved):

jfberry:
Thank you for this.
In general I really like the refactoring that has been done.
Ditto makes my head hurt.

We have a few users testing this, so suggest we leave this for a few days to ensure there are no unexpected problems and then merge

Mygod:
@Fabio1988 I think you are seeing that only because you just transitioned from `main` to this branch. It should get better now? Those are for spawn rerolls and it should not happen very often. But yes we could make spawn reroll logs debug level instead if other people agree.

> Ditto makes my head hurt.

@jfberry Agreed. 👍

PR #69, opened 2023-03-24
Fix a sacrilegious default setting

Mygod:


jfberry:
Happy to take votes from approvers
It was me that set this default in the example because poracle manages this in alternative ways that users generally find less confusing

Mygod:
Well this is PR number 69 (nice) so it will be sadge and not nice if it does not get merged. :joy:

Fabio1988:
as most of us are using false we should merge that. I had also user complaining about those mons. Not useful to have that info in ReactMap :)

PR #71, opened 2023-03-24
Add sentry & pyroscope integrations

lenisko:
Add two integrations, default off.
- Sentry - used to collect errors and metrics https://docs.sentry.io/platforms/go
- Pyroscope - continuous profiling client https://pyroscope.io/docs/golang


Fabio1988:
> Ability to enable Sentry & pyroscope within Golbat. Optional, default off.
> 
> **Note: Those are (mostly) useful for dev purposes!**

Description does not match I guess

external/sentry.go
```diff
@@ -0,0 +1,31 @@
...
+
+		err := sentry.Init(sentry.ClientOptions{
+			Dsn:        targetDsn,
+			Debug:      false,
+			SampleRate: config.Config.Sentry.SampleRate,
+		})
+		if err != nil {
+			log.Errorf("Sentry Init Failed: %s", err)
+		}
+		defer sentry.Flush(2 * time.Second)
```
jfberry, line 29: this won't defer anything (it will execute immediately) - intended?

external/sentry.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/getsentry/sentry-go"
+	log "github.com/sirupsen/logrus"
+	"golbat/config"
+	"time"
+)
+
+const sentryDsn = "https://bc0b5dc992ee486a90faa20b83016351@sentry.mew.is/2"
+
+func InitSentry() {
+	if !config.Config.Sentry.Disable {
```
jfberry, line 13: negative config verbs are horrible. Should be enable.
Should it be enabled by default?

lenisko:
@Fabio1988 updated 👍 I'll add a note about sentry to `config.toml.example` as well

external/sentry.go
```diff
@@ -0,0 +1,31 @@
+package external
+
+import (
+	"github.com/getsentry/sentry-go"
+	log "github.com/sirupsen/logrus"
+	"golbat/config"
+	"time"
+)
+
+const sentryDsn = "https://bc0b5dc992ee486a90faa20b83016351@sentry.mew.is/2"
```
jfberry, line 10: Is this the unownhash shared instance? README should be updated to make clear what is going on.

external/sentry.go
lenisko: You are correct, I'll move it 👍

external/sentry.go
Fabio1988: I agree, change config Variable to ebable, yes it's enabled by default

lenisko: Would definitely speed up error's lookup. But it depends, I can keep it disabled by default. I'll rework to `Enable`

external/sentry.go
jfberry: I am not sure what it is meant to be doing - ie when Flush needs to be called

external/sentry.go
lenisko: Yes it is. Have pending README and `config.toml.example` changes, soon:tm:

external/sentry.go
jfberry: You can make it enable and enable it by default by setting to true before loading config. But this would need to be very clear to users in readme

external/sentry.go
jfberry: Logging to Sentry

The unown has developers have installed a Sentry instance to allow us to track performance and errors in Golbat. This contains no personal data nor anything that identifies your installation (I assume this is true?)
If you wish to disable it include:

```[sentry]
enable=false
```
in your config.toml

lenisko:
```
	// If this flag is enabled, certain personally identifiable information (PII) is added by active integrations.
	// By default, no such data is sent.
	SendDefaultPII bool
```
No PII is sent - https://docs.sentry.io/platforms/go/configuration/options/

external/sentry.go
lenisko: No need. Dropping idea of default DSN.

lenisko:
@jfberry both disabled by default, no default sentry dsn.

jfberry:
This is now stripped right back; I'm happy for the crash logs to be collected by default as long as they can be disabled and this behaviour is explained to users (and its obvious how to turn it off).

lenisko:
@jfberry We'll get there eventually. For now, just want to get this into the base :)

PR #72, opened 2023-03-24
Do not save capture rate in db

TurtIeSocks:
Removes capture rate from the db, since it's entirely useless. Still sends in webhooks though.

Based off of #65 , merge that one first

### DIFF
--- a/decoder/pokemon.go
+++ b/decoder/pokemon.go
@@ -136,8 +136,7 @@ func getPokemonRecord(ctx context.Context, db db.DbDetails, encounterId string) 
 
 	err := db.PokemonDb.GetContext(ctx, &pokemon,
 		"SELECT id, pokemon_id, lat, lon, spawn_id, expire_timestamp, atk_iv, def_iv, sta_iv, iv_inactive, iv, "+
-			"move_1, move_2, gender, form, cp, level, weather, costume, weight, height, size, capture_1, capture_2, "+
-			"capture_3, display_pokemon_id, is_ditto, pokestop_id, updated, first_seen_timestamp, changed, cell_id, "+
+			"move_1, move_2, gender, form, cp, level, weather, costume, weight, height, size, display_pokemon_id, is_ditto, pokestop_id, updated, first_seen_timestamp, changed, cell_id, "+
 			"expire_timestamp_verified, shiny, username, pvp, is_event, seen_type "+
 			"FROM pokemon WHERE id = ?", encounterId)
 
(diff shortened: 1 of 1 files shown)


Fabio1988:
Why useless? Not used in ReactMap? Which impact on storing that information?

Mygod:
These rates differ by account (catch badges). See also: https://github.com/WatWowMap/Chuck/pull/29

TurtIeSocks:
ReactMap does not use this info, due to it being inaccurate for everyone.

jfberry:
Will need rebasing

TurtIeSocks:
Should we also drop the columns and webhook data?

Fabio1988:
As I can think of only Poracle uses it. somebody could reference it in the DTS

https://github.com/KartulUdus/PoracleJS/blob/develop/src/controllers/monster.js#L215

Mygod:
#65 already stopped processing the rates btw. 😝

jfberry:
> > It's right the data parser doesn't record them; the question is whether we should include them in webhooks by calculation magic (or someone should provide a PR to poracle)
> 
> As in each user would save their gym badges to the poracle table to do calculations?

I meant base rates

Mygod:
*Catch badges, not gym badges. :)

TurtIeSocks:
> So for providing the catch rates in front-end we need at least the cpMultiplier, that's part of the pokemon proto. So we would need a possibility to gather that information for every mon
> 
> is the base cpMultiplier in gamemaster available @TurtIeSocks  ?

I don't believe so. But the jsons are available in many places

Fabio1988:
already done in #74 
can we create an issue to not forget this?

ccev:
I'm not really sure why you'd need CPMs from the gamemaster. to calculate a catch rate, you need

- the Pokemon's CPM (provided by Golbat)
- the Pokemon's base capture rate (provided by the gamemaster, in the Mon's pokemon settings)
- certain modifiers (balls, berries, medals, etc) (hardcoded values in the game's backend, can be found on the internet)

Fabio1988:
> I'm not really sure why you'd need CPMs from the gamemaster. to calculate a catch rate, you need
> 
>     * the Pokemon's CPM (provided by Golbat)
> 
>     * the Pokemon's base capture rate (provided by the gamemaster, in the Mon's pokemon settings)
> 
>     * certain modifiers (balls, berries, medals, etc) (hardcoded values in the game's backend, can be found on the internet)

CPM=cpMultiplier ?
yeah it's in golbat, but it varies on the worker, true? if a worker achieves normal catch medal this will be different, no?
#79 we should continue discussion in that gh issue :)

PR #75, opened 2023-03-25
Huge opimizations & cleanup

lenisko:
- **Removed `github.com/google/go-cmp/cmp`**
- Ignore near float differences for `hasChangesSpawnpoint`
- Added docs for refactored methods (and a note to keep them updated after making changes)
- Commented out `log.Traceln`

As ugly as current solution looks like... `cmp.*` been eating a LOOT of resources in Golbat.
> It is intended to only be used in tests, as performance is not a goal and it may panic if it cannot compare the values
https://pkg.go.dev/github.com/google/go-cmp/cmp

Thanks to replacing `cmp.*` Golbat is taking **~2.2 times LESS! CPU time.** Memory improvements also noticed.

~~Considering use of single line conditional expression, any thoughts?~~ Done

2023-03-25_23-04-48-firefox

### DIFF
--- a/decoder/player.go
+++ b/decoder/player.go
@@ -2,8 +2,6 @@ package decoder
 
 import (
 	"database/sql"
-	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/jellydator/ttlcache/v3"
 	log "github.com/sirupsen/logrus"
 	"golbat/db"
@@ -14,6 +12,8 @@ import (
 	"time"
 )
 
+// Player struct. Name is the primary key.
+// REMINDER! Keep hasChangesPlayer updated after making changes
 type Player struct {
 	// Name is the primary key
 	Name               string      `db:"name"`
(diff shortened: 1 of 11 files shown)


decoder/pokemon.go
```diff
@@ -145,11 +145,119 @@ func getPokemonRecord(ctx context.Context, db db.DbDetails, encounterId string)
...
 
+// hasChangesPokemon compares two Pokemon structs
+// Ignore Pvp field
+// Float tolerance for Lat, Lon
 func hasChangesPokemon(old *Pokemon, new *Pokemon) bool {
-	return !cmp.Equal(old, new, cmp.Options{
-		ignoreNearFloats,
-		cmpopts.IgnoreFields(Pokemon{}, "Pvp"),
-	})
+	if old.Id != new.Id {
```
jfberry, line 152: I am glad you included the comment to ignore the pvp field; this characteristic could get lost. 
There are some code generators available for golang which write code for you instead of using reflection; perhaps we should investigate one (though we don't have soooo many objects it is impossible to check)
It would also be possible to sequence these checks to put the most likely changes first to save another few microseconds (lol)

lenisko: Yeah! I've been thinking about adding counters to check which ifs we hit the most, but I will do that once we'll get prometheus :joy:

Yap, tried to make all required notes in parts we might touch in the future, so those methods won't get missed.

lenisko:
I'll rework them to single line conditional expression in a moment (on a walk).

jfberry:
> I'll rework them to single line conditional expression in a moment (on a walk).

Maybe multi-line but single statement :-)

lenisko:
yeah yeah;)

jfberry:
Note new db field from ditto PR, and fix in there where null.Float is now near compared reducing writes

TurtIeSocks (approved):

Mygod:
Hmm it's almost like maybe we should use a ORM or something... 🤔

jfberry:
This kind of reflection based comparison is what orms are filled with of course (and sqlx is a norm, but not a query builder)

Fabio1988 (approved):

decoder/player.go
```diff
@@ -236,15 +236,88 @@ func getPlayerRecord(db db.DbDetails, name string, friendshipId string, friendCo
...
+		old.CaughtFire != new.CaughtFire ||
+		old.CaughtWater != new.CaughtWater ||
+		old.CaughtGrass != new.CaughtGrass ||
+		old.CaughtElectric != new.CaughtElectric ||
+		old.CaughtPsychic != new.CaughtPsychic ||
+		old.CaughtIce != new.CaughtIce ||
+		old.CaughtDragon != new.CaughtDragon ||
+		old.CaughtDark != new.CaughtDark ||
+		old.CaughtFairy != new.CaughtFairy ||
+		!nullFloatAlmostEqual(old.KmWalked, new.KmWalked, 0.001)
```
na-ji, line 320: Good job writing those conditions 🤣

lenisko: Scripted, people make mistakes :)

jfberry:
> You don't want to use ORMs in highly performant applications. That's also off-topic, as this PR is about struct compressions/application layer and not SQL ;-)

There are, however, code generators available which take a structure and bring boilerplate code (like this) from a common source.  
These can be a very good idea when you have lots of models with a very similar structure.

I do have in my mind, though, that we should be evolving away from relational structures and we need to keep some optionality around our approach

lenisko:
I couldn't find any solution you describe for Go.

jfberry:
> I couldn't find any solution you describe for Go.

I found a couple before, probably used terms like "go code generator sql", "go code template engine", ... will look later - but they might not be a good match for our future direction in any case

PR #76, opened 2023-03-26
Update spawnpoints regardless of wild pokemon processing

jfberry:
After being pushed to get the DB performance updates in, some minor issues have arisen. This PR is intended to resolve these.
 * In doing this, have made spawnpoint learning faster
 * Last seen now updates only once per hour to reduce it updating multiple times for the same pokemon appearing
 * Despawn time could bounce around, so will only be updated if it moves by more than a few seconds

### DIFF
--- a/decoder/main.go
+++ b/decoder/main.go
@@ -254,13 +254,13 @@ func UpdateFortBatch(ctx context.Context, db db.DbDetails, p []RawFortData) {
 func UpdatePokemonBatch(ctx context.Context, db db.DbDetails, wildPokemonList []RawWildPokemonData, nearbyPokemonList []RawNearbyPokemonData, mapPokemonList []RawMapPokemonData, username string) {
 
 	for _, wild := range wildPokemonList {
+		encounterId := strconv.FormatUint(wild.Data.EncounterId, 10)
+		pokemonMutex, _ := pokemonStripedMutex.GetLock(encounterId)
+		pokemonMutex.Lock()
+
 		spawnpointUpdateFromWild(ctx, db, wild.Data, int64(wild.Timestamp))
 
 		if config.Config.Tuning.ProcessWilds {
-			encounterId := strconv.FormatUint(wild.Data.EncounterId, 10)
-			pokemonMutex, _ := pokemonStripedMutex.GetLock(encounterId)
-			pokemonMutex.Lock()
-
 			pokemon, err := getOrCreatePokemonRecord(ctx, db, encounterId)
 			if err != nil {
 				log.Errorf("getOrCreatePokemonRecord: %s", err)
@@ -290,9 +290,8 @@ func UpdatePokemonBatch(ctx context.Context, db db.DbDetails, wildPokemonList []
 					}(wild.Data, int64(wild.Cell), int64(wild.Timestamp))
 				}
 			}
-
-			pokemonMutex.Unlock()
 		}
+		pokemonMutex.Unlock()
 	}
 
 	if config.Config.Tuning.ProcessNearby {
--- a/decoder/main.go
+++ b/decoder/main.go
@@ -252,7 +252,6 @@ func UpdateFortBatch(ctx context.Context, db db.DbDetails, p []RawFortData) {
 }
 
 func UpdatePokemonBatch(ctx context.Context, db db.DbDetails, wildPokemonList []RawWildPokemonData, nearbyPokemonList []RawNearbyPokemonData, mapPokemonList []RawMapPokemonData, username string) {
-
 	for _, wild := range wildPokemonList {
 		encounterId := strconv.FormatUint(wild.Data.EncounterId, 10)
 		pokemonMutex, _ := pokemonStripedMutex.GetLock(encounterId)
@@ -266,7 +265,7 @@ func UpdatePokemonBatch(ctx context.Context, db db.DbDetails, wildPokemonList []
 				log.Errorf("getOrCreatePokemonRecord: %s", err)
 			} else {
 				if pokemon == nil || pokemon.isNewRecord() || pokemon.wildSignificantUpdate(wild.Data) {
-					updateTime := time.Now().Unix()
+					updateTime := int64(wild.Timestamp / 1000)
 					go func(wildPokemon *pogo.WildPokemonProto, cellId int64, timestampMs int64) {
 						time.Sleep(15 * time.Second)
 						pokemonMutex, _ := pokemonStripedMutex.GetLock(encounterId)
--- a/decoder/main.go
+++ b/decoder/main.go
@@ -252,8 +252,11 @@ func UpdateFortBatch(ctx context.Context, db db.DbDetails, p []RawFortData) {
 }
 
 func UpdatePokemonBatch(ctx context.Context, db db.DbDetails, wildPokemonList []RawWildPokemonData, nearbyPokemonList []RawNearbyPokemonData, mapPokemonList []RawMapPokemonData, username string) {
-	if config.Config.Tuning.ProcessWilds {
-		for _, wild := range wildPokemonList {
+
+	for _, wild := range wildPokemonList {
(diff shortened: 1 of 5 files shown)


decoder/pokemon.go
```diff
@@ -411,31 +411,27 @@ func (pokemon *Pokemon) isNewRecord() bool {
...
 
-	spawnId, err := strconv.ParseInt(wildPokemon.SpawnPointId, 16, 64)
-	if err != nil {
-		panic(err)
-	}
-
-	// Not sure I like the idea about an object updater loading another object
-
-	pokemon.updateSpawnpointInfo(ctx, db, wildPokemon, spawnId, timestampMs, timestampAccurate)
-	pokemon.SpawnId = null.IntFrom(spawnId)
```
Mygod, line 429: `SpawnId` is no longer getting set? Actually do we even need to persist this in the database? 🤔

decoder/pokemon.go
```diff
@@ -510,7 +506,7 @@ func (pokemon *Pokemon) calculateIv(a int64, d int64, s int64) {
 	pokemon.Iv = null.FloatFrom(float64(a+d+s) / .45)
 }
 
-// wildSignificantUpdate returns true if the wild pokemon is significantly different from the current pokemon and
+// nearbySignificantUpdate returns true if the wild pokemon is significantly different from the current pokemon and
 // should be written.
 func (pokemon *Pokemon) nearbySignificantUpdate(nearbyPokemon *pogo.NearbyPokemonProto) bool {
```
Mygod, line 511: This function is still not used so could remove? Or is this unintended?

decoder/pokemon.go
```diff
@@ -449,7 +445,7 @@ func (pokemon *Pokemon) updateFromWild(ctx context.Context, db db.DbDetails, wil
 	if pokemon.setPokemonDisplay(int16(wildPokemon.Pokemon.PokemonId), wildPokemon.Pokemon.PokemonDisplay) {
 		updateStats(ctx, db, pokemon.Id, stats_statsReset)
 	}
-	pokemon.addWildPokemon(ctx, db, wildPokemon, timestampMs, true)
+	pokemon.addWildPokemon(ctx, db, wildPokemon, timestampMs)
```
Mygod, line 448: This will still call `spawnpointUpdate`. Maybe we can move that part of the functionality into `updatePokemonFromEncounterProto`?

decoder/pokemon.go
jfberry: Set elsewhere now

Mygod (approved):
Didn't test but LGTM.

decoder/main.go
```diff
@@ -178,6 +178,12 @@ var ignoreNearNullFloats = cmp.Comparer(func(x, y null.Float) bool {
 	}
 })
 
+const floatTolerance = 0.000001
+
+func floatAlmostEqual(a, b, tolerance float64) bool {
+	return math.Abs(a-b) < tolerance
+}
+
```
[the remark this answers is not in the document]
jfberry: why?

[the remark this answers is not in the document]
lenisko: It's a slight change, but I think we should return `true` when abs diff between `a` and `b` is equal to the tolerance value.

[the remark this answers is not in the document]
jfberry: this is nonsense because it's a float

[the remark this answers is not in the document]
lenisko: Okay, whatever ;-) I'll update my PR later.

decoder/pokemon.go
```diff
@@ -595,54 +582,31 @@ const SeenType_LureEncounter string = "lure_encounter" // Pokemon has been encou
...
-// a new exact spawnpoint record
-func (pokemon *Pokemon) updateSpawnpointInfo(ctx context.Context, db db.DbDetails, wildPokemon *pogo.WildPokemonProto, spawnId int64, timestampMs int64, timestampAccurate bool) {
-	if wildPokemon.TimeTillHiddenMs <= 90000 && wildPokemon.TimeTillHiddenMs > 0 {
-		expireTimeStamp := (timestampMs + int64(wildPokemon.TimeTillHiddenMs)) / 1000
-		pokemon.ExpireTimestamp = null.IntFrom(expireTimeStamp)
-		pokemon.ExpireTimestampVerified = true
+func (pokemon *Pokemon) updateSpawnpointInfo(ctx context.Context, db db.DbDetails, wildPokemon *pogo.WildPokemonProto, timestampMs int64) {
+	spawnId, err := strconv.ParseInt(wildPokemon.SpawnPointId, 16, 64)
+	if err != nil {
+		panic(err)
```
Fabio1988, line 589: panic is good at this point because GIN recovers? :)

jfberry: It was in the original code from early days. Probably should just return to be honest. We aren’t in a gin call here but a goroutine.

decoder/pokemon.go
```diff
@@ -411,31 +411,27 @@ func (pokemon *Pokemon) isNewRecord() bool {
 	return pokemon.FirstSeenTimestamp == 0
 }
 
-func (pokemon *Pokemon) addWildPokemon(ctx context.Context, db db.DbDetails, wildPokemon *pogo.WildPokemonProto, timestampMs int64, timestampAccurate bool) {
+func (pokemon *Pokemon) addWildPokemon(ctx context.Context, db db.DbDetails, wildPokemon *pogo.WildPokemonProto, timestampMs int64) {
```
Fabio1988, line 414: this method is still called from `updateFromWild` and from `updatePokemonFromEncounterProto`. So also in the `updateFromWild` we do not update the spawnpoint anymore. we only update spawnpoint in `decoder.main` with `spawnpoint.spawnpointUpdateFromWild` ?

decoder/main.go
```diff
@@ -252,18 +258,20 @@ func UpdateFortBatch(ctx context.Context, db db.DbDetails, p []RawFortData) {
...
+		pokemonMutex.Lock()
+
+		spawnpointUpdateFromWild(ctx, db, wild.Data, int64(wild.Timestamp))
 
+		if config.Config.Tuning.ProcessWilds {
 			pokemon, err := getOrCreatePokemonRecord(ctx, db, encounterId)
 			if err != nil {
 				log.Errorf("getOrCreatePokemonRecord: %s", err)
 			} else {
 				if pokemon == nil || pokemon.isNewRecord() || pokemon.wildSignificantUpdate(wild.Data) {
```
Fabio1988, line 273: pokemon can not be `nil` anymore here :)
`getOrCreatePokemonRecord` returns error or a pokemon object (not nullable)

decoder/pokemon.go
jfberry: The idea is spawnpoints are not updated during pokemon processing; they are updated separately. So if we choose to delay, or disable wild pokemon scanning we still get spawnpoints.  
Are you saying that you see updates from within pokemon still?

Fabio1988: you mean updates in DB? did not checked

I only followed the workflow and checked usages of that method in Goland (impossible to see usages in github review itself) :)

Fabio1988: should we add a proper comment to the method call? so we don't confuse us in the future on our own? :)

jfberry: I am unsure what needs to be commented - it was confusing before, and a little bit ugly (that a pokemon update could change a spawnpoint).  Now the two things are separate it is easier to see what causes spawnpoint changes - pokemon now just references this data

decoder/pokemon.go
Mygod: I wrote the panic because something is terribly wrong if this fails.

lenisko (approved):
lgtm

Fabio1988:
As this is much better than actual main we should merge that and work on it later. With all those changes in ditto detection and spawnpoint handling it's better to look on the full picture :)

PR #78, opened 2023-03-28
improve s2 cell update

Fabio1988:
when cell is already cached we don't need to lookup the s2 cell props anymore

### DIFF
--- a/decoder/main.go
+++ b/decoder/main.go
@@ -355,10 +355,8 @@ func UpdateClientWeatherBatch(ctx context.Context, db db.DbDetails, p []RawClien
 	}
 }
 
-func UpdateClientMapS2CellBatch(ctx context.Context, db db.DbDetails, r []uint64) {
-	for _, mapS2CellId := range r {
-		saveS2CellRecord(ctx, db, mapS2CellId)
-	}
+func UpdateClientMapS2CellBatch(ctx context.Context, db db.DbDetails, cellIds []uint64) {
+	saveS2CellRecords(ctx, db, cellIds)
 }
 
 func ClearRemovedForts(ctx context.Context, dbDetails db.DbDetails, mapCells []uint64) {
--- a/decoder/main.go
+++ b/decoder/main.go
@@ -357,9 +357,7 @@ func UpdateClientWeatherBatch(ctx context.Context, db db.DbDetails, p []RawClien
 
 func UpdateClientMapS2CellBatch(ctx context.Context, db db.DbDetails, r []uint64) {
 	for _, mapS2CellId := range r {
-		s2Cell := &S2Cell{}
-		s2Cell.updateS2CellFromClientMapProto(mapS2CellId)
-		saveS2CellRecord(ctx, db, s2Cell)
+		saveS2CellRecord(ctx, db, mapS2CellId)
 	}
 }
 
--- a/decoder/s2cell.go
+++ b/decoder/s2cell.go
@@ -27,24 +27,23 @@ type S2Cell struct {
 //  PRIMARY KEY (`id`)
 //)
 
-func (s2Cell *S2Cell) updateS2CellFromClientMapProto(mapS2CellId uint64) *S2Cell {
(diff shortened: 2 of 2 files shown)


jfberry:
Have we demonstrated this is expensive? I did think I saw something from Len, but can't find it now

jfberry:
Having said that, this is a very simple change that seems 'obviously' better

Fabio1988:
> Have we demonstrated this is expensive? I did think I saw something from Len, but can't find it now

I can't find it anymore either.. But assuming that we used the s2 lib on every seen this might be worth to reduce lib use :)
@len ? :)

jfberry:
That is likely a different issue as these would likely be all new cells. Inserts are generally more expensive than updates. You’d have to run again with a populated database to get a real view probably. 
from this it doesn’t look like the s2 calculation is the culprit

lenisko:
I think we could just bootstrap s2cells in db once, using s2 golang library, after:
- adding new area geofence
- editing area geofence

and run db cleanup after:
- removal of area & geofence

And during normal work bump `updated` column using cellId each x minutes. This would be the most efficient solution. But current proposal is not bad as well.

jfberry:
I can't see how this approach drives anything but complexity. You're effectively doing this now based on an insert/update.  You could reduce the columns in the on duplicate update to only update last updated which might save some floating point comparison in the database.
A cleanup routine could be done already based on last_seen being over a period (assuming this was indexed)
But I doubt, in this table, you'd even notice - the re-ordering of the index would probably be more expensive than leaving the very narrow column (which has no nulls, or variable width fields so is packed nicely on db pages).

lenisko:
> That is likely a different issue as these would likely be all new cells. Inserts are generally more expensive than updates. You’d have to run again with a populated database to get a real view probably. from this it doesn’t look like the s2 calculation is the culprit

I re-read your first comment. No, those are not new inserts. I already had all S2Cells for the whole country before checking pperf results (4 130 094 rows to be precise).

> I can't see how this approach drives anything but complexity. You're effectively doing this now based on an insert/update. You could reduce the columns in the on duplicate update to only update last updated which might save some floating point comparison in the database. A cleanup routine could be done already based on last_seen being over a period (assuming this was indexed) But I doubt, in this table, you'd even notice - the re-ordering of the index would probably be more expensive than leaving the very narrow column (which has no nulls, or variable width fields so is packed nicely on db pages).

Well, yes, that's a bit more complicated, so for now I will leave this topic off and will focus on what we have here.

---

1. I think we can replace the last line with `ON DUPLICATE KEY UPDATE updated=VALUES(updated)`. The rest of S2 cell data is static. We don't have to update it.
https://github.com/UnownHash/Golbat/blob/3de5465fa325d5530410df861e16c743b8d4ecab/decoder/s2cell.go#L49-L53

2. We might as well push all the cells in single sql query (bulk). I believe it's possible to do that even with `ON DUPLICATE KEY UPDATE` in query. For 9x9 that's 81 calls.

jfberry:
> I re-read your first comment. No, those are not new inserts. I already had all S2Cells for the whole country before checking pperf results (4 130 094 rows to be precise).

They may not have been in the cache though...

> Well, yes, that's a bit more complicated, so for now I will leave this topic off and will focus on what we have here.
> 
> 1. I think we can replace the last line with `ON DUPLICATE KEY UPDATE updated=VALUES(updated)`. The rest of S2 cell data is static. We don't have to update it.
>    https://github.com/UnownHash/Golbat/blob/3de5465fa325d5530410df861e16c743b8d4ecab/decoder/s2cell.go#L49-L53

This is what I was suggesting earlier. 

> 2. We might as well push all the cells in single sql query (bulk). I believe it's possible to do that even with `ON DUPLICATE KEY UPDATE` in query. For 9x9 that's 81 calls.

This second idea is probably worth testing, sqlx actually has an easy way to do this (we have used this elsewhere)

lenisko:
@Fabio1988 if you need a hand, ping me 👍

lenisko:
Much better when it comes to CPU usage :)

lenisko (approved):
tested & fast and furious

PR #81, opened 2023-03-29
gc dcm fix

TurtIeSocks:
GC seems to be sending `PokmonGO/1` now instead of `PokmonGO/0`


darthbutcher:
Tested and it corrects the previous error as expected.

lenisko (approved):
Thanks! I was kicked out from GC and couln't follow changes anymore :)

PR #82, opened 2023-03-29
Penultimate Ditto scanning: Ditto scanning even if spawn rerolls

Mygod:
The main highlight of this PR is that when spotlight hour ends (or when a Ditto event begins), more Ditto than before can be immediately detected using IV scanned during spotlight hour (or before the event begins).

Technical details: It should more reliably detect `00/0N>[B0] or 0P>BN` even if spawn rerolls. There's also another extremely rare mode `BN>[0P] or B0>00/0N` that is also newly supported by this PR.

Other minor improvement: If a Ditto rerolls into its own disguise, we can now sometimes catch it and reset it to non-Ditto.

Already tested for the past spotlight hour.


TurtIeSocks (approved):
Ran in production all day without issues.

Fabio1988 (approved):

PR #85, opened 2023-03-29
Update README for DCM compatibility

darthbutcher:
This adds a line about the data source with the correct nomenclaute for DCM

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,12 @@ At this time this is likely to be mostly development discussion.
 
 # Configuration of data source
 
+**For Atlas**
 The data source should be configured to send to Golbat's 
 URL which will be `http://ip:port/raw`
+**For DCM**
+The data source should be configured to send to Golbat's 
+URL which will be `http://ip:port`
 
 # Tuning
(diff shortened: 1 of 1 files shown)


Fabio1988:
How about other data sources?

darthbutcher:
Don't have access to Flygon yet, so did not know that. DCM has two separate fields for endpoints: Backend & Data Endpoint

DCM also has the ability to send raw data to multiple data endpoints, so it would still be good to include instructions for DCM as a separate line, with the caveat that Flygon can also handle this. I will await Flygon before updating this PR.

Are there any other data sources to keep in mind?

lenisko:
He dares to forget about Flygon... Minus points from the first PR!

I believe @Fabio1988 meant that Atlas doesn't use DCM and you can't split the traffic without external tools. Unless something has changed and I'm mistaken.

darthbutcher:
How about a "For DCM" `http://ip:port` and "For Other Data Sources" `http://ip:port/raw` to simply and be inclusive?

PR #90, opened 2023-03-30
Implement scan rules

jfberry:
Scan rules can be added to the configuration. These will be processed in order, first match applies - and allows disabling of processing certain types of game objects.

The scan rules can match the object to a geofence, or use the scanner 'mode' when it is supported by raw senders (looking at you Flygon!)

```toml
[[scan_rules]]
areas = ["MainArea"]
nearby_pokemon = false

[[scan_rules]]
context = ["Scout"]

[[scan_rules]]
pokemon = false
```

Here the main area would not process nearby pokemon.  Messages arriving in 'scout' mode would have everything processed; and the default would not process any pokemon (so outside main area not delivered by the scout service)

* pokemon - any pokemon processing (disables spawnpoints also)
* wild_pokemon - process wild pokemon from GMO
* nearby_pokemon -  process nearby pokemon from GMO
* weather - process weather in GMO
* gyms - process gyms in GMO
* pokestops - process pokestops in GMO
* cells - process cell updates (disabling this also disables automatic fort clearance)

This supercedes the tuning options

### DIFF
--- a/decoder/scanarea.go
+++ b/decoder/scanarea.go
@@ -7,15 +7,16 @@ import (
 )
 
 type ScanParameters struct {
-	ProcessPokemon bool
-	ProcessWild    bool
-	ProcessNearby  bool
-	ProcessWeather bool
-	ProcessForts   bool
-	ProcessCells   bool
+	ProcessPokemon   bool
+	ProcessWild      bool
+	ProcessNearby    bool
+	ProcessWeather   bool
+	ProcessPokestops bool
+	ProcessGyms      bool
+	ProcessCells     bool
 }
 
-func FindScanArea(scanContext string, lat, lon float64) ScanParameters {
+func FindScanConfiguration(scanContext string, lat, lon float64) ScanParameters {
 	var areas []geo.AreaName
 	areaLookedUp := false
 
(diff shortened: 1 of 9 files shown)


lenisko:
Good job! It has a great potential :)

Few things:
- All the checks are run only in `decodeGMO`, what about other places like `UpdatePokemonRecordWithEncounterProto` ?
- Having so many filters, I think it would be great to add `pokestop, gym, invasion`. As for `lure` it might be harder?

jfberry:
This is only for GMOs at this stage. Other messages are specifically called for by the controller by being on a route etc, so didn't think it was worth doing.  A follow-up PR can be made to add capability if required, but I'm not really convinced that it adds anything over and above what we have here.

pokestop vs gym - these are together in forts, as this is the classification in the GMO. 
invasions and lures come for free with pokestops - I see no benefit (performance wise anyway) in stopping their processing. If you don't want them displayed on a map that is easily enough arranged.

lenisko:
> This is only for GMOs at this stage. Other messages are specifically called for by the controller by being on a route etc, so didn't think it was worth doing. A follow-up PR can be made to add capability if required, but I'm not really convinced that it adds anything over and above what we have here.

It would make more sense to apply those filters everywhere. For example, you can't control narrow places for Pokémon scans. With filters, you can exclude "dangerous places" in overall. Here I mean geofence exclusion, but still should apply for all.

> pokestop vs gym - these are together in forts, as this is the classification in the GMO.

I think it would make sense, as they are saved separately. Same as described above.

> invasions and lures come for free with pokestops - I see no benefit (performance wise anyway) in stopping their processing. If you don't want them displayed on a map that is easily enough arranged.

You are correct :) my bad.

jfberry:
Feel free to take this branch forward when I confirm what exists in it already works, I'll let you know when I'm ready to hand it over

lenisko:
No rush 👍 thanks!

jfberry:
I have split fort processing into gyms & pokestops.  I probably will add encounters to pokemon so that there is a pattern for others (rather than it just being GMO), but won't consider any further breakdown

lenisko:
that would be sufficient, thanks!

decoder/scanarea.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+
+		// We have a match
+
+		defaultTrue := func(value *bool) bool {
+			if value == nil {
+				return true
+			}
+			return *value
+		}
```
lenisko, line 53: Move it outside loop

lenisko (changes requested):
One place to correct. Otherwise, looks good and works well 👍

decoder/scanarea.go
jfberry: …because?

lenisko: Why would you want function declaration Inside loop, when you can initialize it once near variables?

jfberry: Because 1) it’s not in a loop that will continue. It’s literally followed by a return statement. 
2) compiler optimisation would mean it is not actually created each loop even if it was. 
3) it is clearer code.

lenisko: 1-2. You are right.
3. Seriously it is?

Before
before

After
after

lenisko (approved):

ccev:
only just seeing this PR and it looks really good. I do have one point I'd like to throw in: I would prefer not differentiating between pokestops and gyms, and combine them to `forts`.

The long-term goal would be to differentiate between them as little as possible in the backdnd. Introducing another place where they're separated now goes against this goal.

Also, i can't think of a use-case where you'd want to disable either gyms or pokestops. Perhaps it might be worth considering adding toggles for fort events like raids/incidents/quests?

Either way, I don't have a strong opinion here. Just wanted to throw it in.

jfberry:
It was originally forts but Len suggested splitting
Given that these are processed and displayed seperately I can see why you may not want to collect pokestop information on a wider basis. 
the model is easily extensible for event types but these come for free in general which is why I didn’t include them in the first draft. I suggest going with what we have now and refining later as we get broader use cases

ccev:
sounds good, let's keep it this way then

Issue #88, opened 2023-03-30
Optimization - geo.MatchGeofences


lenisko:
Another possibility to save CPU time would be to optimize `golbat/geo.MatchGeofences`

It's worth checking the approach with indexing bbox of polygons inside RTree index. Depending on results, we might as well move `MatchGeofences` higher to check all incoming locations.

[Building a Geospatial cache in Go](https://tonyalaribe.medium.com/building-a-geospatial-cache-in-go-ab635777ccb1)

![2023-03-30_13-58-00-firefox](https://user-images.githubusercontent.com/10072920/228830380-207f93a8-0127-42a1-9cfe-266a55ed44e1.png)

![2023-03-30_13-58-22-firefox](https://user-images.githubusercontent.com/10072920/228830371-68f2d56f-0396-4856-a590-3eb895c9f982.png)

lenisko:
#91 results are quite shocking :) Mostly for users with **a loot** geofences, but all of us should see improvement. Even more of CPU time saved! Good job!

**Before**
![2023-03-31_02-09-58-firefox](https://user-images.githubusercontent.com/10072920/228991381-131c7a40-1cb3-46fb-976a-c9f87958aaf1.png)

**After**
![2023-03-31_02-08-19-firefox](https://user-images.githubusercontent.com/10072920/228991179-df74c4da-4035-46f9-92bf-10bf5d00b395.png)

PR #91, opened 2023-03-30
Implement RTrees for faster geographic matching

jfberry:
Golbat isn't fast enough.
MOOOREEE SPEEED.

Issue: #88

### DIFF
--- a/decoder/fort.go
+++ b/decoder/fort.go
@@ -5,7 +5,6 @@ import (
 	"github.com/jellydator/ttlcache/v3"
 	log "github.com/sirupsen/logrus"
 	"golbat/db"
-	"golbat/geo"
 	"golbat/pogo"
 	"golbat/webhooks"
 )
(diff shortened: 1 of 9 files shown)


Mygod:
You can use Max Airstream for more speed, or Agility if +1 is not enough. 😃

ringsaturn:
If need more performance improvements, consider use point-in-polygon from https://github.com/tidwall/geojson, it has a QuadTree/RTree index for polygon’s points https://github.com/tidwall/geojson/issues/23 which is much faster compared with orb’s geojson.

jfberry:
> If need more performance improvements, consider use point-in-polygon from https://github.com/tidwall/geojson, it has a QuadTree/RTree index for polygon’s points [tidwall/geojson#23](https://github.com/tidwall/geojson/issues/23) which is much faster compared with orb’s geojson.

This is a great find, as it would likely have provided the optimisation for us automatically.
I suspect the performance we see now is equivalent as the point in polygon is, I think, the same algorithm and using the rtree to select which polygons to test just as we are now.
Certainly introducing the rtree has made a substantial difference for installations with a large number of polygons and we are now experimenting with it for direct retrieval of pokemon from the cache avoiding sql lookups.

jfberry:
> If need more performance improvements, consider use point-in-polygon from https://github.com/tidwall/geojson, it has a QuadTree/RTree index for polygon’s points [tidwall/geojson#23](https://github.com/tidwall/geojson/issues/23) which is much faster compared with orb’s geojson.

Ps, didn't realise you were the maintainer of tzf - thanks, we love it

PR #97, opened 2023-04-04
Config - add support for environment variables 

na-ji:
Hello,

I added the possibility to configure Golbat from environment variables. To achieve this goal, I used the library [koanf](https://github.com/knadh/koanf). Also, I decided to use Unmarshal, this way the implementation of the new config didn't require updating all our usages. 

The main challenge when implementing this PR was reading slices on environment variables. This required to write a custom envParser, since the default one was overwriting slice members with each others.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -84,21 +84,19 @@ type database struct {
 
 type tuning struct {
 	ExtendedTimeout bool `koanf:"extended_timeout"`
-	ProcessWilds    bool `koanf:"process_wild_pokemon"`
-	ProcessNearby   bool `koanf:"process_nearby_pokemon"`
 }
 
 type scanRule struct {
-	Areas            []string       `toml:"areas"`
-	AreaNames        []geo.AreaName `toml:"-"`
-	ScanContext      []string       `toml:"context"`
-	ProcessPokemon   *bool          `toml:"pokemon"`
-	ProcessWilds     *bool          `toml:"wild_pokemon"`
-	ProcessNearby    *bool          `toml:"nearby_pokemon"`
-	ProcessWeather   *bool          `toml:"weather"`
-	ProcessCells     *bool          `toml:"cells"`
-	ProcessPokestops *bool          `toml:"pokestops"`
-	ProcessGyms      *bool          `toml:"gyms"`
+	Areas            []string       `koanf:"areas"`
+	AreaNames        []geo.AreaName `koanf:"-"`
+	ScanContext      []string       `koanf:"context"`
+	ProcessPokemon   *bool          `koanf:"pokemon"`
+	ProcessWilds     *bool          `koanf:"wild_pokemon"`
+	ProcessNearby    *bool          `koanf:"nearby_pokemon"`
+	ProcessWeather   *bool          `koanf:"weather"`
+	ProcessCells     *bool          `koanf:"cells"`
+	ProcessPokestops *bool          `koanf:"pokestops"`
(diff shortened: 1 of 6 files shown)


config/config.go
```diff
@@ -3,89 +3,89 @@ package config
...
+	Db       string `koanf:"db"`
+	MaxPool  int    `koanf:"max_pool"`
 }
 
 type tuning struct {
-	ExtendedTimeout bool `toml:"extended_timeout"`
-	ProcessWilds    bool `toml:"process_wild_pokemon"`
-	ProcessNearby   bool `toml:"process_nearby_pokemon"`
+	ExtendedTimeout bool `koanf:"extended_timeout"`
+	ProcessWilds    bool `koanf:"process_wild_pokemon"`
```
jfberry, line 87: I think this is removed now

config/config.go
```diff
@@ -3,89 +3,89 @@ package config
...
+	ProcessNearby   bool `koanf:"process_nearby_pokemon"`
 }
 
 type scanRule struct {
```
jfberry, line 91: this doesn't appear to have been converted

docker-compose.yml.example
```diff
@@ -4,17 +4,64 @@ services:
...
+      - GOLBAT.DATABASE.USER=
+      - GOLBAT.DATABASE.PASSWORD=
+      - GOLBAT.DATABASE.ADDRESS=host.docker.internal:3306
+      - GOLBAT.DATABASE.DB=
+
+      - GOLBAT.PVP.ENABLED=true
+      - GOLBAT.PVP.INCLUDE_HUNDOS_UNDER_CAP=false
+      - GOLBAT.PVP.LEVEL_CAPS=50,51
+
+      - GOLBAT.PVP.LEAGUES.0.NAME=little
```
jfberry, line 47: I wonder whether we should take the opportunity to tidy up pvp configuration as these are just trash, especially here in docker config. We could do as a seperate PR (I would be happy to do first and you merge up?) or we could do as part of this one

na-ji: You mean setting default values or removing the config altogether?

config/reader.go
```diff
@@ -1,54 +1,83 @@
...
+		Logging: logging{
+			SaveLogs: false,
+		},
+		Cleanup: cleanup{
+			StatsDays: 7,
+		},
+		Database: database{
+			MaxPool: 100,
+		},
+		Tuning: tuning{
```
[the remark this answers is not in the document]
na-ji: Shit, it's a rebasing fail. I will fix it.

docker-compose.yml.example
jfberry: as per malte's issue discussion (or small variation) - getting rid of the config altogether. The leagues are the leagues, the only real option is whether the 500 lead has the evolution restriction (and the global settings for level caps etc, for which there would be sensible defaults)

jfberry:
Does koanf have same behaviour for the scan rules - ie null if not present but true / false in the bool pointer?
If it does then it looks great

na-ji:
> Does koanf have same behaviour for the scan rules - ie null if not present but true / false in the bool pointer? If it does then it looks great

Yes, I can confirm that the behavior is the same as on main, by default the slice and the boolean pointers are nil. I've tried with the config you provided #90.

jfberry:
Let’s go to market :-)

na-ji:
> Let’s go to market :-)

Alright, thank you 😄

PR #99, opened 2023-04-05
Populate CP on spawn reroll

Mygod:
This PR will allow Golbat to add CP immediately upon seeing the spawn rerolls (if the previously scanned IVs are available). CP was reset in #65 but could not be recalculated due to not having gohbem support. Now this is possible with gohbem v0.10.0.

~~Untested but should be fine?~~ Tested and is fine.

### DIFF
--- a/decoder/pokemon.go
+++ b/decoder/pokemon.go
@@ -244,11 +244,6 @@ func savePokemonRecordAsAtTime(ctx context.Context, db db.DbDetails, pokemon *Po
 		pokemon.FirstSeenTimestamp = now
 	}
 
-	pokemon.Updated = null.IntFrom(now)
-	if oldPokemon == nil || oldPokemon.PokemonId != pokemon.PokemonId || oldPokemon.Cp != pokemon.Cp {
-		pokemon.Changed = now
-	}
-
 	changePvpField := false
 	if ohbem != nil {
 		// Calculating PVP data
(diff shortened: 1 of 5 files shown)


Mygod:
Unfortunately this feature would be turned off by `config.Pvp.Enabled`. This is fine?

lenisko (approved):
Run `go mod tidy` and commit changes (not critical). There are leftovers from last version. Otherwise, looks good.

TurtIeSocks (approved):

decoder/pokemon.go
```diff
@@ -267,13 +262,34 @@ func savePokemonRecordAsAtTime(ctx context.Context, db db.DbDetails, pokemon *Po
 				pokemon.Pvp = null.StringFrom(string(pvpBytes))
 				changePvpField = true
 			}
+			var displayPokemon int64
```
jfberry, line 265: Why is this here in the pvp calculation and not where pokemon id is changed (ie when the wild or nearby is seen?)
Perhaps I am missing something.

The PVP calculation is only in the update so we don't persist it in the in-memory cache as it is so big, but this calculation (aside from using an ohbem function ) appears unrelated

Mygod: Hmm okay.

Mygod:
Okay there is one issue with this PR: A 0P Ditto will have its CP set improperly until rescanned. I think this is an edge case that I am willing to let slip unless others object.

Mygod:
Somehow every PR of mine turned into a Ditto PR in the end.

Mygod:
Merge anyone? 👀

Fabio1988:
Are you on a hurry?! ;)
Did you approve that this fixed the cp = null issue on weatherboost? 

I could not yet, it's holiday here, so keep patient ;)

Mygod:
I don't think that issue has anything to do with my PR. My PR might fix it but that is not its purpose. 😕

PR #105, opened 2023-04-15
New direct-from-cache query experimentation

jfberry:
Golbat/ReactMap in memory Pokemon is finally really good. This branch needs to be accompanied by mem-experiments on ReactMap. 

Add a pokemon key to your Golbat schema in RM config, and set it to the endpoint http://{golbat_api}/api/pokemon/scan 

You no longer need an empty pokemon table/schema, having it as part of your standard golbat schema is fine. 
The only major thing that isn't ready on the reactmap side is area restrictions. Pvp queries are roughly 5-10x faster :ShakingEyes: 

The `in_memory` option in your golbat config is deprecated. The sqlite dependency is removed.

The in memory query works alongside writing to mariadb, but a new option `pokemon_memory_only` can be set which will not read or write to the pokemon table at all.


config/config.go
```diff
@@ -3,21 +3,23 @@ package config
...
-	Tuning    tuning     `koanf:"tuning"`
-	ScanRules []scanRule `koanf:"scan_rules"`
+	Port              int        `koanf:"port"`
+	Webhooks          []webhook  `koanf:"webhooks"`
+	Database          database   `koanf:"database"`
+	Stats             bool       `koanf:"stats"`
+	Logging           logging    `koanf:"logging"`
+	Sentry            sentry     `koanf:"sentry"`
+	Pyroscope         pyroscope  `koanf:"pyroscope"`
+	LegacyInMemory    bool       `koanf:"in_memory"`
```
na-ji, line 13: You might want to rename the config too, according to your description. 
```suggestion
	LegacyInMemory    bool       `koanf:"legacy_in_memory"`
```

jfberry: It's actually removed entirely now... but I have left this piece behind

PR #113, opened 2023-05-30
add raid pokemon alignment

Fabio1988:
followed same structure as in RDM https://github.com/RealDeviceMap/RealDeviceMap/pull/523


jfberry:
lgtm (assuming you have tested it!)

Mygod:
This is technically a redundant field since raid level also indirectly shows alignment? (Up until now.)

Fabio1988:
Untested yet.. 
Using raid level might not be the right choice. This alignment is also used in wild pokemon, who knows, maybe Niantic let them spawn in the future ;)

Fabio1988:
works as intended. column is properly filled with values, no crashes, all fine :) 
lgtm if nobody is against that additional column?!

PR #117, opened 2023-06-12
Pokestop api endpoints

TurtIeSocks:
- adds `POST` `/api/pokestop-positions` to get pokestops within a specified geofence
- adds `GET` `/api/pokestops/id/:fort_id` to get the specific PokeStop

### DIFF
--- a/db/pokestop.go
+++ b/db/pokestop.go
@@ -3,14 +3,16 @@ package db
 import (
 	"context"
 	"database/sql"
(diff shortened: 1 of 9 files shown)


Fabio1988 (approved):

jfberry:
looks good to me; we still have inconsistent policy around re-ordering imports.  I think goland and vscode can be configured in the same way we just need to decide

PR #119, opened 2023-06-14
keep device map clean

TurtIeSocks:
- keep the device map trimmed by setting a configurable ttl value in the cache


jfberry:
This is a great addition. Any particular reason you needed to use cron rather than just running hourly based on a timer?
eg https://github.com/UnownHash/Golbat/blob/main/stats.go#L216 quest expiry runs every hour and one minute
(trying to avoid keeping things from running at the same time by different offsets in this expiry strategy)

jfberry:
P.S. You don't really need to take care that this one doesn't coincide with anything else as it's all in memory and will be near instant.
It could also be changed to use the LRU cache like other elements instead; the one we use can be iterated and is thread-safe and performance is not a big concern given the small size of this data structure

TurtIeSocks:
Only reason I chose cron was experience with it from another project. I missed that we're already using LRU elsewhere in the codebase, I'll dig into that tonight and swap over. Thanks!

jfberry:
Cron vs timer -- we use timer elsewhere.
But LRU is what we use for all other caches, so that is probably the thing to swap to

jfberry:
just noticed go.mod looks a little mixed up version wise
can you update deps and tidy up go modules?

```bash
go get -u ./...
go mod tidy
```

TurtIeSocks:
Done

PR #123, opened 2023-06-26
Update protos to 0.275.x

jfberry:
As description


Mygod (approved):
Nice but it probably should be in a separate crate if other projects also need to use it.

jfberry:
Other projects will need to get the vbase direct from source, we cannot distribute beyond here

PR #127, opened 2023-07-17
send fort_update webhook only if image path changed

Fabio1988:
nia changed the image url schemas recently - from http to https


jfberry:
Looks good, merge when tested

PR #128, opened 2023-07-20
adapt checks for fort changes

Fabio1988:
this prevents webhooks send if e.g. Description changes from nil -> ""

Ready to review, Already checked out and running with that changes... Next changes incoming in about 4 hours hopefully, then I can check behavior

### DIFF
--- a/decoder/fort.go
+++ b/decoder/fort.go
@@ -162,7 +162,7 @@ func CreateFortWebHooks(old *FortWebhook, new *FortWebhook, change FortChange) {
 		}
 
 		// Check if ImageUrl has changed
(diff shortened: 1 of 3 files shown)


lenisko (approved):
Hopefully we won't get `"" => NULL` :)
Otherwise, looks good! 

Thank you!

jfberry:
Did this actually happen? I can’t conceive of when it could?

lenisko:
happened to me, once though.

Fabio1988:
Tested since days, all good, notification webhooks are good. System stable

jfberry (approved):

PR #133, opened 2023-07-27
Add routes support

na-ji:
Golbat will now decode routes received through the `GetRoutesOutProto`. The routes are saved in a new table called `route`. 

This PR closes issue #129.


sql/19_route.up.sql
```diff
@@ -0,0 +1,21 @@
...
+    `start_poi`          varchar(35)         NOT NULL,
+    `end_poi`            varchar(35)         NOT NULL,
+    `image`              varchar(200)        NOT NULL,
+    `image_border_color` varchar(10)         NOT NULL,
+    `reversible`         tinyint(1) unsigned NOT NULL,
+    `tags`               text DEFAULT NULL,
+    `type`               tinyint unsigned    NOT NULL,
+    `updated`            int unsigned        NOT NULL,
+    `version`            int unsigned        NOT NULL,
+    `waypoints`          text                NOT NULL,
```
Mygod, line 17: Should `waypoints` and `tags` use `json` type instead of `text`? (I guess other tables use `text` too. Just wondering.)

sql/19_route.up.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE `route`
+(
+    `id`                 varchar(35)         NOT NULL,
+    `name`               varchar(100)        NOT NULL,
+    `description`        varchar(200)        NOT NULL,
```
Mygod, line 5: Are these limits correct? A UTF-8 string can have different length vs byte length. I don't think we need to be that aggressive?

sql/19_route.up.sql
```diff
@@ -0,0 +1,21 @@
...
+    `distance_meters`    int unsigned        NOT NULL,
+    `duration_seconds`   int unsigned        NOT NULL,
+    `start_poi`          varchar(35)         NOT NULL,
+    `end_poi`            varchar(35)         NOT NULL,
+    `image`              varchar(200)        NOT NULL,
+    `image_border_color` varchar(10)         NOT NULL,
```
Mygod, line 11: I think this can be an `int unsigned` to save space (9 bytes vs 4 bytes).

sql/19_route.up.sql
jfberry: Unfortunately not all sql drivers support the json type. Since we're not querying directly into the json this shouldn't matter.

na-ji: Also, on Mariadb the JSON type is just an alias for `LONGTEXT`. So it doesn't really matter.

sql/19_route.up.sql
na-ji: Actually, the in game max length for the name is 49 characters and the description is 239 characters. So which is suitable in SQL?

lenisko: UTF8 can take up to 4 bytes per character. Maybe make it 255 for both? MySQL will take only required space + 1 byte (up to 255 including).

https://dev.mysql.com/doc/refman/8.0/en/char.html

Mygod: So for name we can just use `varchar(255)`, and for description maybe just `varchar(65535)` or something?

Mygod: Another option is to not save descriptions at all.

jfberry: > So for name we can just use `varchar(255)`, and for description maybe just `varchar(65535)` or something?

You can't have strings that long in mariadb; and you want to avoid the combination of all fields going over a database page (well it won't let you anyway). The general accepted practice is to make them as big as you need and not much bigger ;-)

lenisko: > So for name we can just use `varchar(255)`, and for description maybe just `varchar(65535)` or something?

lol no. Just up to 255 you have lower cost of storing data, not stored data alone.

Mygod: Yeah I guess saving description up to 255 chars is also a reasonable alternative to not storing at all. :)

jfberry:
No indexes yet for route retrieval

decoder/routes.go
```diff
@@ -0,0 +1,205 @@
...
+	EndLat           float64 `db:"end_lat"`
+	EndLon           float64 `db:"end_lon"`
+	Image            string  `db:"image"`
+	ImageBorderColor string  `db:"image_border_color"`
+	Reversible       bool    `db:"reversible"`
+	StartFortId      string  `db:"start_fort_id"`
+	StartImage       string  `db:"start_image"`
+	StartLat         float64 `db:"start_lat"`
+	StartLon         float64 `db:"start_lon"`
+	Tags             *string `db:"tags"`
```
jfberry, line 30: Should this be null.String ?

decoder/routes.go
```diff
@@ -0,0 +1,205 @@
...
+			WHERE id = :id`,
+			route,
+		)
+
+		if err != nil {
+			return fmt.Errorf("update route error %w", err)
+		}
+	}
+
+	routeCache.Set(route.Id, *route, ttlcache.DefaultTTL)
```
jfberry, line 153: Not sure if we need a webhook?
Can be later, but I'm not sure anything would alert on a new route

sql/19_route.up.sql
```diff
@@ -0,0 +1,29 @@
...
+    `type`               tinyint unsigned    NOT NULL,
+    `updated`            int unsigned        NOT NULL,
+    `version`            int unsigned        NOT NULL,
+    `waypoints`          text                NOT NULL,
+    PRIMARY KEY (`id`),
+    KEY `ix_coords_start` (`start_lat`, `start_lon`),
+    KEY `ix_coords_end` (`end_lat`, `end_lon`)
+) ENGINE = InnoDB
+  DEFAULT CHARSET = utf8mb4
+  COLLATE = utf8mb4_unicode_ci;
```
jfberry, line 29: CHARSET/COLLATE does not match other tables preventing join
`) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;`

decoder/routes.go
na-ji: Shit, I knew there was a way to handle this

decoder/routes.go
na-ji: I do think that webhook would be useful, the same as for fort updates. But that should be something for another PR.

Fabio1988 (approved):

decoder/routes.go
```diff
@@ -0,0 +1,204 @@
...
+	route.EndImage = sharedRouteProto.GetEndPoi().GetImageUrl()
+	route.EndLat = sharedRouteProto.GetEndPoi().GetAnchor().GetLatDegrees()
+	route.EndLon = sharedRouteProto.GetEndPoi().GetAnchor().GetLngDegrees()
+	route.Image = sharedRouteProto.GetImage().GetImageUrl()
+	route.ImageBorderColor = sharedRouteProto.GetImage().GetBorderColorHex()
+	route.Reversible = sharedRouteProto.GetReversible()
+	route.StartFortId = sharedRouteProto.GetStartPoi().GetAnchor().GetFortId()
+	route.StartImage = sharedRouteProto.GetStartPoi().GetImageUrl()
+	route.StartLat = sharedRouteProto.GetStartPoi().GetAnchor().GetLatDegrees()
+	route.StartLon = sharedRouteProto.GetStartPoi().GetAnchor().GetLngDegrees()
```
Mygod, line 172: I don't think we actually need to preserve start/end latlon? I think these are just the latlon for the pokestop/gym and not actually part of the route?

Mygod: I just checked a route and can confirm that these lat lon are just stop/gym lat lon. We could use them to update the gym/stop table lat lon. Not sure why this should be in the table though?

Fabio1988: It's uses by RM to fetch routes within few.. changing that to fort id would only cause useless joins

Mygod: Hmm, we could change it to the actual start/end latlon from waypoints, instead of the start/end poi?

Fabio1988: Why having a useless dependency on something that buggy implemented from nia? Why should we call for a fort when we have everything in route proto?

Nobody knows what's happening with routes if start/end fort is moved ingame, maybe route disappears ;)

But route and forts are totally different calls, keep them different as well, don't merge useless information

Mygod: I think it's clear that these lat lon are just intended to be tied to the poi itself instead of a bug as you described. A route's start/end has to be tied to a poi; it has to be near it but doesn't necessarily need to be its exact start/end point; it's part of the design?

Fabio1988: But then you answered your question on your own? It does not necessarily to be exact start/end point. So why modifying the route and using fort lat/lon? Useless logic, keep it as it is ;)

Mygod: For ReactMap, you ideally want to actually query the start/end lat lon, instead of the lat lon of the start/end poi, which is not actually the starting/ending point?

jfberry: The routes start and end at a poi; you have to set these when you record a route - and when you follow it is the POI which triggers the start and end.
Have you followed/created a route yet?  The route does not even appear until you go to the start POI.

jfberry (approved):

decoder/routes.go
Mygod: Pretty sure you can see the route even away from the POI. Are you sure you need to stand on the precise point of the POI to start the route?

jfberry: > Pretty sure you can see the route even away from the POI. Are you sure you need to stand on the precise point of the POI to start the route?

When I have completed the routes the POI is indicated with a marker pointing to it, and it has a distance countdown to the POI, and when you get close enough to the POI the route lights up.  You have to be within a normal pokemon-style distance of the POI to begin it.
Similarly with the route ending, the POI is highlighted as the endpoint that you have to get close to in order to trigger route completion.

When you create routes the first thing you do is nominate the start and end POI which is used during the recording process.  Recording can start as you get to the POI, and when you reach the final POI it pops up and asks you to confirm your route is completed.

ccev: routes are too fresh of a feature to draw the assumption that images and coordinates are always connected to a PokeStop. the way they're sent in the proto leaves them the option to make a custom start location. drawing assumptions like this and not strictly following the protos is a big mistake we've made multiple times in the past.

Mygod: I'm not sure if I'm following the discussion. I feel like we are talking about different things. Let me maybe submit a PR and explain there.

Mygod: https://github.com/UnownHash/Golbat/pull/135

PR #114, opened 2023-05-31
Support for expert filter for in-memory Pokemon

Mygod:
Rebased on main thanks to squashing. 😡

PR #134, opened 2023-07-29
DNF Pokemon filters

Mygod:
This is the next-gen Pokemon filter designed to both replace the current default+additional+pvp filter (which has complicated and hacky logic and is ReactMap centric) and supersede the user-unfriendly expert filter from #114, while also being super performant. Functionality wise it is identical to expert filters (and absolutely destroys the old filter), and for most use cases it should be as efficient as expert filters. At the same time, the code (and the logic) is more readable/accessible than the other two filters. Currently the old filters are still kept for backwards compatibility reasons but moving forward they should probably be removed.

Implementation details: A new json array field `dnf` is added to `ApiPokemonFilter`, which will override and disable old set of filters (similar to #114). Matching logic is follows.

0. All rulesets are categorized by Pokemon-form, or a default category if it is unset.
1. If the current Pokemon matches a specific Pokemon-form, that ruleset will be used, otherwise the default ruleset is used.
2. A row is returned in the result if any rule in the ruleset is matched.
3. A rule is matched if every single set standard is satisfied by the current row. (Therefore, overall it is an OR of AND's, and hence a Disjunctive Normal Form.)

A consequence of this logic is that an empty ruleset will match nothing. To include everything, one could, for example, have a rule matching IV in range `[-1, 100]`.

Change to old filters: per-Pokemon filter will override global filters, otherwise it is not possible to for example always hide a certain species of Pokemon (without creating a filter for every Pokemon-form). This also allows for 2x performance for filtering.

@jfberry @TurtIeSocks Comments please. (Also needs corresponding support from ReactMap.) Closes #114.


jfberry:
I am on my mobile so no chance to review in detail. But in general I think this is the right idea. 

I would make this an entirely new api so we can test in parallel in final form rather than having a combined call. That also makes it easier to delete the old one. 

I am also interested whether we can support the api over grpc as that should give us some marshalling performance but since it requires a protobuf based input we should perhaps leave that until later (as we would need to tie down request and response format)


Finally for any testing we should compile with the go_json flag (as per makefile) as the json marshaller was one of the weak spots in previous benchmarking

Mygod:
This PR is meant to be just a start. We can do further performance optimizations down the road. Let's not cram too many things into this.

Re combined call: I don't see any perceivable difference for me at least. I can help with removing old code in the future.

jfberry:
It would be better to have the separate api calls now, guaranteeing backward compatibility and providing for a cleaner implementation of the new model. 
This would be the conventional way to provide a v2 api - the v1 deprecation is straight forward and the v1 api is exactly backwards compatible all the time it is still in use. 

I will have more time to look later.

Mygod:
Ready to review/test/benchmark now. This commit/branch also removes the old filter from the codebase: https://github.com/Mygod/Golbat/tree/killoldfilter

jfberry:
On my phone but I don’t see the second api in routes?
Both APIs should be available in this version through different endpoints. The original one should be unaltered. 
will look later when at a laptop

Mygod:
No need to make a separate API endpoint if it's just for testing?

jfberry:
Surely testing will involve seeing the results are the same between the two endpoints and comparing performance etc?

Mygod:
If your JSON contains a dnf field, it will use the new API; otherwise it uses the old API.

jfberry:
Having just picked through this in more detail, I think this really should be two totally independent functions and routes so that we can just remove the old one; and the existing one is totally untouched by any new code.
That way we can have them living side by side for however long we need to; test them head to head; and remove the old one very simply.
This is standard practice for a v2 api and saves a lot of the confusion in the code which exists because you have tried to make one function do both apis. Harder to validate it gives the same results as the old one in the old instance.

Secondly, for the new DNF there is an opportunity to think about the way that the api is called.
The pokemon ID_form string as a map source is poor for performance as it requires a string conversion for every entry in the loop (this was something >10% of execution time last time I measured); either at the api level or (since you seem to be building up a different list internally) you can make the key a structure with pokemon id/form id.   I'd be in favour of doing this at the api level so it is clear how to structure this.
The other question we should ask ourselves is whether the two-int array is the right way or a more explicit min/max - this doesn't make much of a difference but does force the validation of length to save doing this each loop.

Are there any other changes to naming that we should make whilst we are breaking the api?

jfberry:
Finally, in the hot loop, where you are looping through an array it is faster to do it by array indexing rather than range (probably worth a few %age points)

Mygod:
I think for everyone's sanity sake, it is better to just remove the old one right now and stop procrastinating any further. If we need any optimizations, we just do it on this instead of comparing it with an alternative approach filled with insanity.

If I'm not mistaken, this API is only used by RM only, and will only be used by RM? Why are we stressed about this like this is a big deal? We can always throw further optimizations at it down the route.

If you really want to optimize, you could also look into how using `||` vs `|` impacts performance thanks to branch predictors in modern CPUs. Maybe we would like to test this across many different architectures of CPUs to see how to best implement this? /s

jfberry:
People upgrade at different times which is why we will support both APIs for a period. 
the optimisations I mentioned were shown in sentry which we will use again

Mygod:
With that said, feel free to change the specific implementation details (API specifics or optimizations alike). Pushing directly is welcome. :)

jfberry:
I thought you were the one that wanted this change? I’m trying to help you get it to a stage where I will merge it.

Mygod:
You can help by pushing onto this branch directly. I don't really care about how you want to do this so it would be more efficient if you just push your suggested changes directly without going through me.

jfberry:
If you would like to wait until I have the time to do so, I will. I don’t expect it will be for a few weeks

Mygod:
I don't really think the API specifics matter since it's RM-facing only, and since the JSON parsing is inevitable, making it into an JSON object instead of a string only increases the overhead. On the other hand, the optimization of removing stringifying the pair can be done without changing the API.

jfberry:
We already have other implementers. And I would also like to create a grpc version of the api after this is complete; and try to avoid future breaking changes. So I was asking to consider the right form of the external api since the form as a string externally is non obvious (and based on RM internal
Details). 

But I agree that if it is the right thing to do it can be converted internally to the most appropriate model for processing.

Mygod:
Added commit 9a6675b3cac281e1a539521447b406683b7647cc as the optimization but it does not change the API. The `for range` optimization is probably too aggressive for too little gain. I'll leave the rest to you.

TurtIeSocks (approved):
💯💯💯💯

jfberry:
I'll separate the routes when I get round to it, or you can - we need the functions and api definitions to be segregate what went before (and will eventually be deleted) to the new api, which will also have a grpc entry point.

jfberry:
Had some time this morning; I have separated the routes which makes the changes a lot clearer and allows for easy removal of the v1 api at the appropriate time.
Looking at the code again it may well be that we need to break up the implementation slightly when adding the grpc version 
 (since the inputs and outputs will be generated protobufs) but the core of it can be re-used

jfberry:
I also note that this implementation - a per pokemon filter not also matching against the 'global' filters means that it does support reactmap logic without sending an 'or' match for every single pokemon of the gloabl filters.  I assume that this is "your" intention, but not sure it matches with reactmap's implementation.

Mygod:
> Change to old filters: per-Pokemon filter will override global filters, otherwise it is not possible to for example always hide a certain species of Pokemon (without creating a filter for every Pokemon-form). This also allows for 2x performance for filtering.

What's the point of writing anything any more? 🤷‍♂️

jfberry:
> > Change to old filters: per-Pokemon filter will override global filters, otherwise it is not possible to for example always hide a certain species of Pokemon (without creating a filter for every Pokemon-form). This also allows for 2x performance for filtering.
> 
> What's the point of writing anything any more? 🤷‍♂️

Yes, I was pointing out that this change requires you to add the extra global filters for each form filter; and not making this change - as you point out - would require you not using a global filter when you wanted to hide certain pokemon.
Of course, RM doesn't allow you to hide pokemon that don't match the global filter which is why I was pointing this out to make sure that turtle was comfortable as the main api user.

I personally think the approach you have taken is correct. I also think we should match against a Pokemon form 0 if exact form filter doesn't exist (and stop there), but I didn't make this change arbitrarily.

Mygod:
> I also think we should match against a Pokemon form 0 if exact form filter doesn't exist (and stop there), but I didn't make this change arbitrarily.

I think this is already handled in ReactMap but I could be wrong.

jfberry:
> > I also think we should match against a Pokemon form 0 if exact form filter doesn't exist (and stop there), but I didn't make this change arbitrarily.
> 
> I think this is already handled in ReactMap but I could be wrong.

In this case I was thinking about other api users that may not have full game data to hand and wish to match on a Pokémon by id without knowing the current form.   Reactmap likely would not use or require this extension.

Mygod:
What other API users do you have in mind? I don't think Golbat has access to masterfile so wouldn't be able to know the default form id. (Always checking form 0 is a bad idea.)

jfberry:
I know of 3 others users of the api
Master file is available to golbat through ohbemgo, but I was thinking of checking Pokémon id, form=0 if exact match not found before global to aid the simple user case when form doesn’t matter to the consumer. 
Golbat would have the choice to optimise by adding all form alternatives from master file or taking the extra step. 
Why do you think this is a bad idea?

Mygod:
We should simply not overload a simple query with hidden features (or potentially future bugs). Are you trying to support a wildcard query like (pokemon id, any form)? If so, we should make it explicit by having form=-1 or something. Is there a specific use case for this?

jfberry:
It wouldn’t be intended to be a hidden feature, but a fully documented one. 
yes, the intention would be to a provide simple way of querying by pokemon id for any form. 
the use case is as I described above - api users who do not care to distinguish between forms or do not have access to the game master

jfberry:
Perhaps the pokemon filter entry should look like this:
```json
{
  "pokemon": [{ "id": xxx, "form": yyy }, {}],
  "iv": { "min": x, "max": y },
```
etc
then if id is null, it is the global match; if id present but form null it is pokemon id only; and if form present exact match
For min/max values if min or max not specified we could use -1 / 65k

This would be largely self-documenting, remove "magic" strings and arrays and also make this mirror a form which converts to protobuf nicely.
Internally we would simplify this structure into something that is easy to match again (broadly what we have now I would guess).

jfberry:
In starting to build the grpc request body I noticed that each Pvp league is now an 'and' section rather than the previous special case or (which makes sense).  Given this, there is no longer any point in having it in a 'sub-filter' and it should be brought up to align with the rest of the filter types.

If you are happy with both of these changes, I can make them - I will publish the grpc request for comment separately.

jfberry:
possible grpc request body (results not yet built) which would align with proposed json

```proto
syntax = "proto3";

option go_package = "github.com/unownhash/golbat/grpc";

package pokemon_api;

// Interface exported by the server.
service Pokemon {
  rpc Search (PokemonSearchRequest) returns (PokemonSearchResponse) {}
}

message PokemonSearchRequest {
  float min_lat = 1;
  float min_lon = 2;
  float max_lat = 3;
  float max_lon = 4;
  int32 limit = 6;
  repeated PokemonDnf filters = 7;
}

message PokemonDnf  {
  optional PokemonId pokemon = 1;
  optional RangeMinMax Iv       =2;
  optional RangeMinMax AtkIv     =3;
  optional RangeMinMax DefIv     =4;
  optional RangeMinMax StaIv   =5;
  optional RangeMinMax Level =6;
  optional RangeMinMax Cp     =7;
  optional int32 Gender   =8;
  optional RangeMinMax PvpLittleRanking =9;
  optional RangeMinMax PvpGreatRanking = 10;
  optional RangeMinMax PvpUltraRanking = 11;
}

message PokemonId {
  int32 id = 1;
  optional int32 form = 2
}

message RangeMinMax {
  int32 min = 1;
  int32 max = 2;
}

message PokemonSearchResponse {
  int64 results = 1;
}
```

Mygod:
`optional repeated PokemonId pokemon = 1;`?

jfberry:
I assumed no PokemonId record would be global, but equally it could be non optional and the id field becomes optional (this may be better)

jfberry:
Since no further comments, I have changed the json version to line up with this grpc version (in the dnfgrpc branch). 
I'll do some manual testing of this when I get the chance.
I will build a grpc response for the pokemon object also

Mygod:
Updated https://github.com/WatWowMap/ReactMap/pull/809 with the latest changes to the API. Also fixed some small issues regarding edge cases.

TurtIeSocks (approved):

jfberry:
In the branch indicated I have changed the api model to reflect the grpc layout naming I proposed. Since you had not commented I assumed you were ok with this. 

I will take these fixes across and open that as a PR to move discussion there

Mygod:
Sure. Meanwhile can this be merged? :eyes:

jfberry:
The api is changed in the branch as discussed.

Mygod:
Well I don't know what you want me to do since your branch is not ready to use yet.

Mygod:
I took a [closer look](https://github.com/Mygod/Golbat/compare/dnfpokemonfilter...UnownHash:Golbat:dnfgrpc). It appears that functionality-wise your branch is identical. If it is something as trivial as changing implementation specifics (grpc or json, minmax or array, api name, whatever), there's no need to go through me. Do whatever you want since it's your repo. At the end of the day I only care about functionality. 🤷

jfberry:
I value your opinion and it is based on your original work. I will open it as a PR in its own right, bring across your fixes here. 
If you are happy with the proposed tweaks to the api, then I would be very happy to receive any examples of head-to-head benchmarking between old and new APIs as it might take me some time to be able to find a large enough system test to make this meaningful. 
My expectation, assuming that there are no unexpected performance issues that need to be understood is that it will then be possible to merge while the change streaming capability using grpc is built out

Mygod:
Sure. Do whatever you want as long as no functionalities are cut. 👍

Mygod:
Superseded by #158.

PR #135, opened 2023-07-31
Fix start/end latlon

Mygod:
Currently in #133, `(Start|End)(Lat|Lon)` is set to the coordinates of the PoIs instead of the coordinates of the actual route. Since these four columns are used for indexing, the actual starting points and ending points should be used instead.

## What evidence do I have for justifying this change?

1. As can be seen in gameplay (for example [this one](https://www.youtube.com/watch?v=fv5CYL6hvb4&t=177s)), the starting point is only near the PoI instead of being exactly on it. 
2. In the proto, these information are under `start_poi` and `end_poi`, which means that they are information belonging to that poi. For example, there are three *separate* images for each of the route, the start poi, and the end poi.
3. Only lat lon is provided for the forts while elevation is also provided for the route coordinates, further indicating that the forts are not part of the route.
4. In [the dump](https://github.com/UnownHash/Golbat/issues/129#issuecomment-1651560267), poi lat lon has truncated precisions whereas waypoint lat lon has full accuracy. This indicates that they come from different data source.
5. I have personally looked at a route and confirmed that the lat lon reported there belong exclusively to the PoIs and not the route.

### DIFF
--- a/decoder/routes.go
+++ b/decoder/routes.go
@@ -161,20 +161,21 @@ func (route *Route) updateFromSharedRouteProto(sharedRouteProto *pogo.SharedRout
 	route.DurationSeconds = sharedRouteProto.GetRouteDurationSeconds()
 	route.EndFortId = sharedRouteProto.GetEndPoi().GetAnchor().GetFortId()
 	route.EndImage = sharedRouteProto.GetEndPoi().GetImageUrl()
-	route.EndLat = sharedRouteProto.GetEndPoi().GetAnchor().GetLatDegrees()
-	route.EndLon = sharedRouteProto.GetEndPoi().GetAnchor().GetLngDegrees()
+	waypoints := sharedRouteProto.GetWaypoints()
+	route.EndLat = waypoints[len(waypoints)-1].LatDegrees
+	route.EndLon = waypoints[len(waypoints)-1].LngDegrees
 	route.Image = sharedRouteProto.GetImage().GetImageUrl()
 	route.ImageBorderColor = sharedRouteProto.GetImage().GetBorderColorHex()
 	route.Reversible = sharedRouteProto.GetReversible()
 	route.StartFortId = sharedRouteProto.GetStartPoi().GetAnchor().GetFortId()
 	route.StartImage = sharedRouteProto.GetStartPoi().GetImageUrl()
-	route.StartLat = sharedRouteProto.GetStartPoi().GetAnchor().GetLatDegrees()
-	route.StartLon = sharedRouteProto.GetStartPoi().GetAnchor().GetLngDegrees()
+	route.StartLat = waypoints[0].LatDegrees
+	route.StartLon = waypoints[0].LngDegrees
 	route.Type = int8(sharedRouteProto.GetType())
 	route.Updated = time.Now().Unix()
 	route.Version = sharedRouteProto.GetVersion()
-	waypoints, _ := json.Marshal(sharedRouteProto.GetWaypoints())
-	route.Waypoints = string(waypoints)
+	waypointsJson, _ := json.Marshal(waypoints)
+	route.Waypoints = string(waypointsJson)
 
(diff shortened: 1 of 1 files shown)


jfberry:
The issue is if reactmap draws the start and end (which it shouldn't) and that issue should be reported there.
These are the lat,lon of the forts is to discover the activating start/end POI and to be able to use this to find routes through the index.
It should not be used to start and end drawing of the route, which is clearly just the waypoints.
It is also clear the end POI triggers the end of the route regardless of whether you have reached the route end through the waypoints or not.

jfberry:
> > Since these four columns are used for indexing, the actual starting points and ending points should be used instead.

They are used to discover the start/end forts to decide whether a route should be shown on a map. Since this is where the route activation points are.
They should not be used as part of drawing.

jfberry:
> If it's not part of the drawing, then it shouldn't be indexed? Regardless, I think the actual start/end latlon should be indexed since that's what's actually being drawn on the map instead of some invisible anchor.

An index is used to locate a record.
The lat,lon in the record is the trigger POI, and the completion POI.
This provides a convenient way to find the route record avoid a join on fort id, nothing more.

For drawing the route, all the waypoints you need are in the json.

I am perfectly happy with the current modelling - if reactmap is displaying these co-ordinates as part of the route trail I would agree that is an incorrect drawing implementation; but it doesn't relate to whether this is a valid way to efficiently discover a record _in the database_ which is all this is about.

Mygod:
Okay but regardless which way we go with, we are not going to join the tables during ReactMap queries? What use case would this *actually* impact?

jfberry:
> Okay but regardless which way we go with, we are not going to join the tables during ReactMap queries? What use case would this _actually_ impact?

The duplication of fort lat,lon here is so that a route can be found to be displayed on the screen of a map if the starting/ending POI is also on the screen - preventing 2x join against the pokestop table to receive the same.
It's not intended to be part of the route.

Mygod:
I disagree. A route should be on the screen if and only if the actual starting/ending point is on the screen. The forts that the route is attached to is irrelevant for displaying purposes (and all other purposes as far as I'm concerned).

(Irrelevant to this but you'll also need to be joining the gym table if you're doing that since a gym could also be attached to routes.)

jfberry:
Even more joins avoided.
The start/end is nearby the POI and these are the activation points. It's really clear in game that these are where you actually go to start and end the route and the waypoints are what you follow in between this (although the display of the route in blue in game is just the waypoints).
Reactmap or other maps can choose to draw their start symbol at waypoint[0] or this lat,lon [of the fort], or by matching to the fort id.

I am not in favour of making this change. I do think it's a shame that the naming is not clearer in identifying that it is a fort lat/lon but I'm not going to get bent out of shape over this.
I am prepared to let Naji consider and make the ultimate decision as it's his PR.

Mygod:
You avoided what? The whole point is that there are no more joins needed one way or the other.

jfberry:
> You avoided what? The whole point is that there are no joins needed one way or the other.

I am coming from the position that the activation POIs are important to be the badge indicators of the start and end of the route, as they are in game. And that the locations of these are therefore the important ones for fast route discovery.
If they aren't in this table and indexed, you would need to join to both fort tables to get the locations.

Mygod:
Can you give a single example use case that a join is actually needed? As I've already argued above, the following use case makes no sense.

> a route can be found to be displayed on the screen of a map if the starting/ending POI is also on the screen

@Fabio1988 If you think it's useless then there's no harm in merging.

jfberry:
In addition, it is the forts presence in the cell asked for which determines whether the route will be shown to you in game.

Fabio1988:
> @Fabio1988 If you think it's useless then there's no harm in merging, so you shouldn't take a stance.

No, I agree on Jabes, no poibt in merging... RM is drawing the lines, not Golbat

Mygod:
Okay just checked myself. I guess you are right.

Issue #125, opened 2023-07-06
Contest Processing


ccev:
it might be cool to have proper contest support. here's the needed data:

GMO -> forts -> PokestopIncidentDisplayProto
```
pokestop_displays {
    incident_id "-1601456624153831644"
    incident_start_ms 1688626800000
    incident_expiration_ms 1688666400001
    incident_display_type IncidentDisplayType.INCIDENT_DISPLAY_TYPE_INCIDENT_CONTEST:9
}
```

GetContestDataOutProto
```
METHOD_GET_CONTEST_DATA | 2105

GetContestDataProto {
    fort_id "ed8a635917594ef2bc25b36993b2dbf0.16"
}

GetContestDataOutProto {
    status Status.SUCCESS:1
    
    contest_incident {
        
        contests {
            contest_id "ed8a635917594ef2bc25b36993b2dbf0.16-CUSTOM-MAX-POKEMON_SIZE-1688630400000"
            
            focus {
                
                pokemon {
                    pokedex_id HoloPokemonId.SQUIRTLE:7
                }
            }
            
            metric {
                pokemon_metric ContestPokemonMetric.POKEMON_SIZE:1
                ranking_standard ContestRankingStandard.MAX:2
            }
            
            schedule {
                
                contest_cycle {
                    start_time_ms 1688630400000
                    end_time_ms 1688666400000
                    contest_occurrence ContestOccurrence.CUSTOM:8
                    custom_cycle_warmup_duration_ms 3600000
                    custom_cycle_cooldown_duration_ms 1
                }
            }
        }
    }
}

```

GetPokemonSizeContestEntryProto
```
METHOD_GET_POKEMON_SIZE_CONTEST_ENTRY | 2104

GetPokemonSizeContestEntryProto {
    contest_id "ed8a635917594ef2bc25b36993b2dbf0.16-CUSTOM-MAX-POKEMON_SIZE-1688630400000"
    entry_count 2147483647
    
    contest_metric {
        pokemon_metric ContestPokemonMetric.POKEMON_SIZE:1
        ranking_standard ContestRankingStandard.MAX:2
    }
}

GetPokemonSizeContestEntryOutProto {
    status Status.SUCCESS:1
    total_entries 13
    
    contest_entries {
        pokedex_id HoloPokemonId.SQUIRTLE:7
        score 900.9230982058893
        rank 1
        trainer_name "Inzeckt"
        team Team.TEAM_BLUE:1
        pokemon_id 16024912515536508900
[...]

PR #136, opened 2023-08-02
Showcase decode

jfberry:
** NOTE: ATLAS DOES NOT CURRENTLY SUPPORT THIS PROTO **

This PR captures my work around decoding the showcase detail protocols.  My intention is to provide a minimum viable decode, rather than a future proof one.  I will work with map makers, etc to get the right level of information and no more.  This is because there is still a lot we haven

Some observations:
* Max number of entries is not in the proto
* Size is the only contest type
* incident id is not in any of the contest protocols, meaning any attempt to store this data on the incident row would require searching for it which may be fragile (or incident may not be present yet)
* Some showcases have >100 entries, and storing these seems to be of low value
* The request is needed for `GetPokemonSizeContestEntry` since the only way to get the 'fortId' is from the contest id in the request
* All the information about the contest can be obtained from GetPokemonSizeContestEntry aside from the end time and schedule information (which are only in `GetContestData`)

Implementation:
* Decode GetPokemonData to store the pokemon id, metric, and end time. This can be done just with the response 
* From GetPokemonSizeContestEntry (requires request)
  * Decode top 3 ranking pokemon and store these including form, costume etc (for fancy display) in a json field.

This would creates just four extra fields on pokestop:

```golang
ShowcasePokemon            null.Int    `db:"showcase_pokemon_id" json:"showcase_pokemon_id"`
ShowcaseRankingStandard    null.Int    `db:"showcase_ranking_standard" json:"showcase_ranking_standard"`
ShowcaseExpiry             null.Int    `db:"showcase_expiry" json:"showcase_expiry"`
ShowcaseRankings           null.String `db:"showcase_rankings" json:"showcase_rankings"`
```

### DIFF
--- a/decoder/pokestop.go
+++ b/decoder/pokestop.go
@@ -58,7 +58,10 @@ type Pokestop struct {
 	AlternativeQuestTitle      null.String `db:"alternative_quest_title" json:"alternative_quest_title"`
 	AlternativeQuestExpiry     null.Int    `db:"alternative_quest_expiry" json:"alternative_quest_expiry"`
 	Description                null.String `db:"description" json:"description"`
-
+	ShowcasePokemon            null.Int    `db:"showcase_pokemon_id" json:"showcase_pokemon_id"`
+	ShowcaseRankingStandard    null.Int    `db:"showcase_ranking_standard" json:"showcase_ranking_standard"`
+	ShowcaseExpiry             null.Int    `db:"showcase_expiry" json:"showcase_expiry"`
+	ShowcaseRankings           null.String `db:"showcase_rankings" json:"showcase_rankings"`
 	//`id` varchar(35) NOT NULL,
 	//`lat` double(18,14) NOT NULL,
 	//`lon` double(18,14) NOT NULL,
@@ -169,7 +172,11 @@ func hasChangesPokestop(old *Pokestop, new *Pokestop) bool {
 		old.AlternativeQuestExpiry != new.AlternativeQuestExpiry ||
 		old.Description != new.Description ||
 		!floatAlmostEqual(old.Lat, new.Lat, floatTolerance) ||
-		!floatAlmostEqual(old.Lon, new.Lon, floatTolerance)
+		!floatAlmostEqual(old.Lon, new.Lon, floatTolerance) ||
+		old.ShowcaseRankingStandard != new.ShowcaseRankingStandard ||
+		old.ShowcasePokemon != new.ShowcasePokemon ||
+		old.ShowcaseRankings != new.ShowcaseRankings ||
+		old.ShowcaseExpiry != new.ShowcaseExpiry
 }
 
 var LureTime int64 = 1800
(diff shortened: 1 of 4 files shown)


clburlison:
> From GetPokemonSizeContestEntry -
Decode pokemon focus from request, and store this in pokemon table

This is the only part that confuses me. Did you mean the Pokestop table?

Fabio1988:
Thanks for extracting the information:
I agree on all points. as we don't get information about incident id we might need to store that information on pokestop itself, or a new table.

> There may be an argument to add the end time to the pokestop as well so you don't have to cross reference to invasion (requiring the GetContestData decode).
> 
> On reflection, I don't think storing the full entries, or extracting the top-3 pokemon into scores directly on the table, provides many benefits at this stage (although open for debate)

When we store without any time on pokestop table, the columns will have data also when the contest is not enabled or when contests are switching stops after season.  
Frontend should only show the showcase if the user is filtering for them. not sure if RM can work like that on two different tables? @TurtIeSocks

Fabio1988:
i had an exhausting discussion with my community of show-case mapping. Hunting XXL with the map is very easy, so those already have an advantage. The points are still missing, so someone playing the game does not see show-case member points before attending. showing those points would give every map user an unfair advantage.

many of those XXL hunter do not attend the showcase until 30 minutes before the end of the showcase. So there is a lot happening in the last 30 minutes. There is also a [bug](https://niantic.helpshift.com/hc/en/6-pokemon-go/faq/4167-pokemon-cannot-be-switched-in-a-pokestop-showcase-once-the-contest-limit-is-reached/) were you can't switch pokemon if you reached contest limit of 3.

So yes it is nice to show the contest itself, but points not. would be nice to keep points behind a toggle in any Frontend / Webhook listener supporting showcases.

Me as admin would definitely disable points for my community, because it leads to an unfair advantage for map only users. Because this information is not even visible ingame, until you take part of the showcase (only 2 more showcases left after)

Mygod:
The bug was resolved already. I agree that it's nice to have a toggle but since there is a 3-mon limit for each account, I don't think the possibility of cheating is that concerning.

By the way, you can also see the points of a contest, unjoin the contest by switching a different stop showcase to the Mon you joined, then repeat. This means that a legit player could also mass hunt showcases in principle.

Mygod:
Well it sounds like that it's more important to have a toggle for displaying/filtering sizes than showcase points. :)

jfberry:
@TurtIeSocks this now writes to the database, so ready for map support when you get a moment
I chose to put the expiry date on the pokestop table as this could be long standing (after the incident has disappeared) and represents the end time of the last showcase on the stop

Mygod:
Well as discussed in #125, since the contest scores and the number of entries are not separate columns, it would be inefficient for map to filter by those. (A very weak argument could be made that this way it would be harder to be toxic.)

> Max number of entries is not in the proto
https://github.com/UnownHash/Golbat/issues/125#issuecomment-1625359437

PR #140, opened 2023-08-04
Fix QuestTitle in Webhook

ReuschelCGN:
fix for QuestTitle so Poracle can decode the Quest_task


jfberry:
Thank you for spotting this

PR #141, opened 2023-08-05
Skip processing unpublished routes

Mygod:
According to community reports, some rejected routes were showing on map. This would stop Golbat from saving them into db.


Fabio1988:
Should we also add a log if a route was parsed if not published? I guess all of us have some ... But we could also truncate and restart golbat as well

jfberry:
Do we know what happens if a route is withdrawn ?

Mygod:
According to protos, a route can be in four states: under review, published, decayed (maybe it means no review in a reasonable timeframe?), rejected. It doesn't seem like there is a state for withdrawn. Do you mean withdrawn by player or Niantic?

jfberry:
I mean by Niantic (you can report routes). It’d be nice if somehow we noticed they were gone. 
This may have to be like forts (ie we get a cell covered but no results) or they may move to a different state

Fabio1988:
I agree, right now we all have routes in our DB which are not published, because Niantic somehow decided to send this information to the client. I wonder about the status "decayed".. maybe this is used when player are reporting already existing routes?

I would prefer a check if a route is already stored in DB, and not published, then pls either log or delete the route in general

Mygod:
That would slow down performance though. Just truncate the table and you will be good. 😎

Fabio1988:
How to remove the ones reported by niantic ;)
Should I truncate daily or what's your idea? ;)

jfberry:
Does the table have last_seen ? We can delete after 24 hrs of not seen for removals

Mygod:
No need to delete. Just have RM filter out 24 hrs of not seen. ;)

We also don't know what would happen if Niantic withdraws a route. Presumably they just wouldn't be sent.

Fabio1988 (approved):

Fabio1988:
there is no possibility to filter out 24 hrs of not seen routes. But right now I guess this PR is the right solution to prevent hidden routes stored in DB

main.go
```diff
@@ -494,6 +494,9 @@ func decodeGetRoutes(payload []byte) string {
 
 	for _, routeMapCell := range getRoutesOutProto.GetRouteMapCell() {
 		for _, route := range routeMapCell.GetRoute() {
+			if route.RouteSubmissionStatus.Status != pogo.RouteSubmissionStatus_PUBLISHED {
+				continue
```
Fabio1988, line 498: I tested this change and added an additional log line on line 498... we never get into the "Route not Published".... even I have a route which is not published yet. so that change does not skip processing unpublished routes

Fabio1988:
I truncated my route table and restarted golbat... instead of 283 it populated 269 routes only. so Niantic might have fixed that issue, no need for that change :)

Mygod:
There's also no reason not to merge this change though. It could change at any time. 🤔

main.go
Mygod: Hi Niantic!

Fabio1988:
I abused this branch to add some more things:
- update route at least every 15 minutes if seen (initial logic to delete routes in the future if deleted from game)
- modify column waypoints to mediumtext (there were reports that the column was too short for big routes)
- add a warning if we find a unpublished route in protos

jfberry:
Seems like a good way forward and we can start seeing if routes are being withdrawn and whether the status is being used. 
lgtm

PR #142, opened 2023-08-06
Align routes log messages to same format.

Cronick:
Title mostly.
Changed log.warnf messages so they are same format, and modified the generic "POST /retrieve/" to match the right URLs.


jfberry:
Thanks for noticing, we changed the api routes during development and obviously missed the warning messages.

Cronick:
Had to start somewhere to look into the file-cache thing, and why not just update it while looking at it 😄

PR #145, opened 2023-08-14
Update go modules - go 1.21 compatibility

jfberry:
as description


jfberry:
@na-ji docker needs updating because go.mod has a format change

jfberry:
Don't know if this also means that this checkin will mean project no longer builds with earlier go?

na-ji:
> @na-ji docker needs updating because go.mod has a format change

It looks using the 1.21 tag was enough

PR #146, opened 2023-08-14
Add showcase and gen9 badges to player table

na-ji:
Hello,

I've added two new columns on the player table. However:
- ~Gen9 is not enabled yet, but will probably be in september.~ It's now enabled
- Showcase badge seems to only be for max size showcases. They might add more badges later. 
- Route badges that were unused are still unused now. So I removed the comments. 
- Route stats are in another proto which isn't available in the public profile.

### DIFF
--- a/decoder/player.go
+++ b/decoder/player.go
@@ -68,6 +68,7 @@ type Player struct {
 	TinyPokemonCaught  null.Int    `db:"tiny_pokemon_caught"`
 	JumboPokemonCaught null.Int    `db:"jumbo_pokemon_caught"`
 	Vivillon           null.Int    `db:"vivillon"`
+	MaxSizeFirstPlace  null.Int    `db:"showcase_max_size_first_place"`
 	DexGen1            null.Int    `db:"dex_gen1"`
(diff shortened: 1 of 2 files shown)


na-ji:
Timing is right since Niantic just enabled the gen 9 badge 

image

Fabio1988 (approved):

jfberry:
Migration 21 already in use, so this needs changing to 22; otherwise happy [or we could merge legacy stats at 22 and make this 23 - happy for either order]

Fabio1988 (changes requested):
@na-ji pls rename migration due to merge of https://github.com/UnownHash/Golbat/pull/143

Use `23_xxxx`

Fabio1988 (approved):
I assume you tested already? Code looks good, migration now as well, lgtm

na-ji:
> I assume you tested already? Code looks good, migration now as well, lgtm

Yes, I now have both the gen 9 values and the showcase. 

image

jfberry (approved):

PR #148, opened 2023-08-18
Keep track of empty cells in fort clearing

Fabio1988:
**This fixes the issue with updating s2 cells even if the whole gmo was empty**

Don't process falsy empty cells, try to keep track of it

- only process s2 cells with content
- if a s2 cell is empty 5 times in a row we consider them as empty and process it
- only clean forts in cells which are processed

### DIFF
--- a/decoder/s2cell.go
+++ b/decoder/s2cell.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"github.com/golang/geo/s2"
 	"github.com/jellydator/ttlcache/v3"
+	"github.com/puzpuzpuz/xsync/v2"
 	log "github.com/sirupsen/logrus"
 	"golbat/db"
 	"gopkg.in/guregu/null.v4"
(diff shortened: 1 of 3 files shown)


decoder/s2cell.go
```diff
@@ -74,3 +75,35 @@ func saveS2CellRecords(ctx context.Context, db db.DbDetails, cellIds []uint64) {
...
+func (tracker *EmptyCellTracker) ResetCount(cellId uint64) {
+	tracker.Data.Delete(cellId)
+}
+
+func (tracker *EmptyCellTracker) ShouldConsiderEmpty(cellId uint64) bool {
+	count, exists := tracker.Data.Load(cellId)
+	if !exists {
+		return false
+	}
+	return count >= 5
```
jfberry, line 108: As I said before, I am concerned that you will consistently get empty cells at the edge of eg 9x9 when outside of max radius; then along comes another type of scan (eg quests vs raid) and now it is suddenly full.
I am not sure how you diagnose this permanently instead of guessing at numbers though (unless you never do a clear when a cell is empty)

main.go
```diff
@@ -658,14 +669,17 @@ func decodeGMO(ctx context.Context, protoData *ProtoData, scanParameters decoder
 	if scanParameters.ProcessCells {
 		decoder.UpdateClientMapS2CellBatch(ctx, dbDetails, newMapCells)
 		if scanParameters.ProcessGyms || scanParameters.ProcessPokestops {
-			if !(len(newMapPokemon) == 0 && len(newNearbyPokemon) == 0 && len(newForts) == 0) {
-				decoder.ClearRemovedForts(ctx, dbDetails, newMapCells)
-			}
+			decoder.ClearRemovedForts(ctx, dbDetails, newMapCells)
```
jfberry, line 672: This could be in a goroutine with a separate context as it seems that this is a more expensive operation where people are seeing GMO timeouts?

decoder/s2cell.go
Fabio1988: This is quite impossible to handle in Golbat, we would have to consider the scanContext, that's not used by any public controller :/

In this case empty really means -> no forts and no mons.
It can also be that a cell does not have any content in one RAW, but in the next there is -> this emptyCell check is more for that reasons to prevent processing empty GMOs

jfberry: If this eliminates false removes, fine -- but I suspect it won't and we'll have to move to a time based approach to catch the 1->0 removals reliably. It's worth a go but since you've increased it I imagine it's not actually working

lenisko:
Attaching problematic case.

9x9 jump is in this location `50.247691,18.982348` (green circle on the right)
Problematic POI is under location `50.254856,18.970153` (center, above mouse pointer)

This is a case where I get rapidly notifications about creation/removal of this POI, with latest changes only with creation (so something has changed).

Most likely, 9x9 are too big for forts scans in some rare occasions?

2023-08-17_12-20-51-chrome

Fabio1988:
follow-up PR

PR #149, opened 2023-08-21
Don't override GMO tth with an encounter tth

jfberry:
As per description
This may improve things if your server time is inaccurate, or your processing is slow and you get an encounter after the pokemon was due to hide.


decoder/pokemon.go
```diff
@@ -644,6 +644,19 @@ func (pokemon *Pokemon) updateSpawnpointInfo(ctx context.Context, db db.DbDetail
...
+			log.Infof("Encounter [END] %s: Spawnpoint %d Timestamp %d ExpireTimestampVerified: %t ExpireTimestamp: %v", pokemon.Id, spawnId, timestampMs, pokemon.ExpireTimestampVerified, pokemon.ExpireTimestamp)
+
+			return
+		} else {
+			timestampMs = time.Now().UnixMilli() // Use current timestamp, accepting that this may be inaccurate
+		}
+	}
+
 	pokemon.SpawnId = null.IntFrom(spawnId)
 	pokemon.ExpireTimestampVerified = false
```
Fabio1988, line 661: this is used by the frontend to consider the tth as confirmed or not.

if the spawnpoint has valid tth we are setting `expireTimestampVerified` to true on line 675.
I guess it can happen that sometimes we encounter before Map Pokemon is processed, and then we would return on line 654 with `expireTimestampVerified=false` :)

jfberry: The ExpireTimestamp.Valid is checking if it is non null (ie there is a value)
timestampMs == 0 is only for an encounter - this is passed through from the GMO if there is one (assuming atlas does not mess with the GMO)

Fabio1988: But this means that we already set the ExpireTimestamp, but this is happening on line 674?! So might be a race condition?

jfberry: Remember we striped lock against pokemon which should mean all updates are discrete. This is why I have added extra logging

decoder/pokemon.go
```diff
@@ -655,6 +655,19 @@ func (pokemon *Pokemon) updateSpawnpointInfo(ctx context.Context, db db.DbDetail
 		panic(err)
 	}
 
+	log.Infof("Encounter [START] %s: Spawnpoint %d Timestamp %d ExpireTimestampVerified: %t ExpireTimestamp: %v", pokemon.Id, spawnId, timestampMs, pokemon.ExpireTimestampVerified, pokemon.ExpireTimestamp)
+
+	if timestampMs == 0 {
```
Fabio1988, line 660: I checked usages and this timestamp might be never 0. On encounter we pass now() :)

jfberry: In which case that needs to be changed (I thought I did) so that this logic can work correctly.

PR #151, opened 2023-08-25
Added options to set save_logs settings

Cronick:
10 files only last for about 5 hours with debug-mode on ~50 devices.
So added option to set it to XX files, so for larger setups it's possible to extend the log history keeped.


Fabio1988:
Would be nice to follow the naming of the module :)
Can you pls add MaxSize and MaxAge as well?

- MaxBackup
- MaxAge
- MaxSize

jfberry:
And defaults when config not specified
(thanks for your contribution!)

Fabio1988 (approved):

logsetup.go
```diff
@@ -16,10 +17,10 @@ func SetupLogger(logLevel log.Level, fileLoggingEnabled bool) {
 	lumberjackLogger = &lumberjack.Logger{
 		// Log file absolute path, os agnostic
 		Filename:   filepath.ToSlash("logs/golbat.log"),
-		MaxSize:    50, // MB
-		MaxBackups: 10,
-		MaxAge:     30,   // days
-		Compress:   true, // disabled by default
+		MaxSize:    config.Config.Logging.MaxSize, // MB
```
Fabio1988, line 20: I wonder if we should do the config loading when calling `SetupLogger()` and add them as params or if we should keep it like that? @jfberry

jfberry: You can argue for a seperation of concerns where these parameters are chosen outside, and config is not imported to logsetup.

PR #153, opened 2023-09-16
Support PogoDroid lat/lng/quests_held

comstud:
PogoDroid sends "quests_held" as a list with quest_ids, instead of "have_ar". This list can be scanned for an AR task to come up with "have_ar".

Additionally, the parsing of the lat/lng from PogoDroid was missing. This adds the lat/lng parsing, which fixes the device location tracking such that PD devices will now show up in ReactMap.

### DIFF
--- a/routes.go
+++ b/routes.go
@@ -7,6 +7,7 @@ import (
 	"golbat/config"
 	"golbat/decoder"
 	"golbat/geo"
+	"golbat/pogo"
 	"io"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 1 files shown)


jfberry:
I am happy with these changes. Since this does not impact the standard path, and I don't run pogodroid - please just confirm that you have tested this; and if so I will merge it.

comstud:
Yep, I was running this exact code all day yesterday, and a less cleaned up version of it for quite some time before then.

comstud:
I still owe the MR to bump the LimitReader size, however. I'll get that up.

PR #154, opened 2023-09-16
Do not copy logs and cache into docker image

comstud:
The cache/logs shouldn't be copied back into new docker image builds, which is what happens when you attach these as volumes, run Golbat for a while, and then re-build the image. This was shoving ~175M of my logs to the docker daemon during build and copying into the image!

I think it only makes sense for people to attach these as volumes at runtime, and doing anything with them at image build time could just end up being an expensive build time no-op.

But even if not, if people do not attach these as volumes, they are always going to be empty on the host side such that there's nothing to copy into the image. But we still need to create the directories.

If you've compiled outside of docker, this also stops the sending of the 130M golbat binary through to the docker daemon when you next build, also.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,3 +2,6 @@
 .idea
 Dockerfile
 config.toml
+cache/
+logs/
+golbat
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,13 +6,14 @@ COPY . .
 
 RUN go mod download
 RUN CGO_ENABLED=0 go build -tags go_json -o /go/bin/golbat
+RUN mkdir /empty-dir
 
 # Now copy it into our base image.
 FROM gcr.io/distroless/static-debian11 as runner
 COPY --from=build /go/bin/golbat /golbat/
+COPY --from=build /empty-dir /golbat/cache
+COPY --from=build /empty-dir /golbat/logs
 COPY /sql /golbat/sql
-COPY /cache /golbat/cache
-COPY /logs /golbat/logs
 
 WORKDIR /golbat
 CMD ["./golbat"]


Fabio1988 (approved):

na-ji:
I don't agree with this change for those reasons:
- Without those folders, Golbat will crash. The advantage of the current implementation is that they are optionals. Since the image is based on a distroless image, mkdir isn't available so that's why I used copy.
- Those folders are always empty on the CI so it's not really an issue for this kind of environment.

I do agree however that the content of the folders shouldn't be copied. If you can make it so the folders are still created by default without their content, that would be better.

jfberry:
I guess the difference here is the CI build vs a local build… nice to find a solution to both!

comstud:
> I don't agree with this change for those reasons:
> 
> * Without those folders, Golbat will crash. The advantage of the current implementation is that they are optionals. Since the image is based on a distroless image, mkdir isn't available so that's why I used copy.
> * Those folders are always empty on the CI so it's not really an issue for this kind of environment.
> 
> I do agree however that the content of the folders shouldn't be copied. If you can make it so the folders are still created by default without their content, that would be better.

Oh, oops, yes. I did mean to create them still, obviously. :)

comstud:
Thnx for the feedback. Please re-review. I've added the empty directory creations. That was just an oversight. While doing this, I also noticed my 130M golbat binary passed through to the docker daemon for no reason, since it is going to be rebuilt in there. So I added that to .dockerignore, also.

na-ji (approved):
Smart solution, thanks!

comstud:
np, ty!

PR #155, opened 2023-09-16
Refactor webhook collector/sender

comstud:
The tl;dr is that this PR refactors the webhooks collector/sender and
adds support for "pokemon_iv" and "pokemon_no_iv". There are some
noteable improvements like reusing http clients for keep-alive,
converting map lookups to slice lookups, removal of hook type
string matching, filtering multiple webhooks in parallel, etc.

This also adds graceful shutdown, ensuring that webhooks are flushed
before exiting.

Details:

* Convert the webhook type constants to uint8 from strings

* Convert the webhook code into a webhookSender struct, injecting and
validating the config, while converting the config.webhook objects into
local webhook objects that contain a slice of WebhookType vs slice of strings.

* The webhook collection is now an array, 1 entry for each type. The
collection is indexed by WebhookType. No more map here.

* When it is time to send webhooks, the current collection is handed off
to 1 goroutine per webhook url where the collection is filtered and sent.
This means that each webhook url is filtering concurrently. Previously,
only the sending was happening concurrently.

* Also adds support for only webhooking pokemon with IVs ("pokemon_iv")
and pokemon without IVs ("pokemon_no_iv").

* Start to rely less upon the global Config. Inject it, instead, as a
dependency. Testing code will be easier without relying upon it. For
example, the webhookSender just requires you pass anything that has a
GetWebhooks() method that returns a slice of config.Webhook.

* Set up a sync.WaitGroup and context that will be used to handle graceful
shutdown. Watch for SIGINT, SIGTERM and cancel the context when received (unix).

* main() now puts the webhook sender into a goroutine itself, so that
  cancelling the previously mentioned context will shut it down.

* wait for things to shutdown, flush webhooks. If http server takes
  longer than 5 seconds, it will close any open connections from clients
  and stop not so gracefully, but flushing webhooks will still occur.


config/config.go
```diff
@@ -40,6 +86,54 @@ type webhook struct {
 	Types     []string       `koanf:"types"`
 	Areas     []string       `koanf:"areas"`
 	AreaNames []geo.AreaName `koanf:"-"`
+
+	typesWanted []WebhookType
```
Fabio1988, line 90: do we need koanf setting for this one?

comstud: Nope. It's private. koanf can't see it.

config/config.go
```diff
@@ -1,6 +1,52 @@
...
+)
+
+type WebhookType uint8
+
+// PayloadType returns the string to be used in the webhook payload for this type.
+func (t WebhookType) PayloadType() string {
+	return webhookTypeToPayloadType[t]
+}
+
+const (
```
jfberry, line 15: It's a shame these are here and not in the webhook module - feels like it's in the wrong place and the master definition should be in the webhook sender.

webhooks/collector.go
```diff
@@ -23,148 +22,75 @@ func (webhookList *WebhookList) AddItem(item WebhookMessage) {
...
 func collectHooks() []WebhookQueue {
 	collectionAccess.Lock()
 	currentCollection := webhookCollections
-	SetMaps()
+	initCollections()
 	collectionAccess.Unlock()
 
 	var destinations []WebhookQueue
 
 	for _, hook := range config.Config.Webhooks {
```
jfberry, line 76: It is difficult to tell by reading; can you confirm that this still works where is no list specified at all (ie null before), in which case all webhooks are sent?

jfberry (commented):
Thank you for your contribution.
I am happy with the general changes and breaking pokemon into iv/non-iv.
I probably wouldn't want to see small breakouts where it could easily be the responsibility of the receiver just to throw things away

webhooks/collector.go
comstud: Good question. Yes. In webhook.initAfterLoad(), if there are no types defined in the config, it copies all possibilities into .typesWanted such that they'll be returned by the call to TypesWanted().

config/config.go
comstud: Yes, it is a shame, I think.

I don't really like this type and constants being in 'config' either. But it looks like 'config' is supposed to be validating the config. There's evidence from existing panic()s. So, I added that in here. 'config' needs to know the possible values in order to do that.

So, we can put them in 'webhooks' where they belong, but then 'config' would need to import 'webhooks' to know the possible values to allow (unless we just want to duplicate them all). But it's a shame that 'config' cannot import 'webhooks'. :) 'webhooks' imports 'config'.

I have a few thoughts on how to redo this. I'm going to sleep soon and will revisit this later and come back with something better. Suggest ideas in the meantime, if you want.

Thnx for calling this out.

comstud:
I have not tested this yet... and in fact, I am going to add some tests to this PR.

main.go
```diff
@@ -32,7 +36,30 @@ var db *sqlx.DB
...
+	defer cancel_fn()
+
+	// cancel context upon being signaled to exit
+	wg.Add(1)
+	go func() {
+		defer cancel_fn()
+		defer wg.Done()
+
+		sig_ch := make(chan os.Signal, 1)
+		signal.Notify(sig_ch, syscall.SIGINT, syscall.SIGTERM)
```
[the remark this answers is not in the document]
jfberry: Whilst the target is not windows, there may be developers and others running on windows. Perhaps under WSL.
I don't have a windows machine so I can't tell what the implications of this decision will be in respect to signals.

What is the intention of this handler?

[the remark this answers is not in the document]
comstud: Graceful shutdown. Flush webhooks before exiting, for one.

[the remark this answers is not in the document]
comstud: I'll fix it so this won't bomb a windows compile. It's easy and it's the right thing to do.

comstud:
This is ready for re-review. I am running this code. Example from a shutdown:

```
INFO 2023-09-24 18:21:59 received signal 'terminated'
INFO 2023-09-24 18:21:59 Starting shutdown...
INFO 2023-09-24 18:21:59 http server is shutdown, waiting for other go routines to exit...
INFO 2023-09-24 18:21:59 go routines have exited, flushing webhooks now...
INFO 2023-09-24 18:21:59 There are 0 webhooks to send to http://poracle-XXXX:3030
INFO 2023-09-24 18:21:59 There are 3 webhooks to send to http://poracle-YYYY:3030
INFO 2023-09-24 18:21:59 Golbat exiting!
```

comstud:
(These last pushes were just to cover more cases in tests)

Fabio1988:
I assume this is again ready to review? Are you running those changes? :)

jfberry:
> I assume this is again ready to review? Are you running those changes? :)

I believe it is ready to review, I was planning on doing so today but real work got in the way. If you or anyone else would like to test it

comstud:
yeah, it's ready. I've been running it since my comment on 9/24. The tests I added should cover it all, too.

config/config.go
```diff
@@ -21,6 +24,15 @@ type configDefinition struct {
 	ScanRules         []scanRule `koanf:"scan_rules"`
 }
 
+func (configDefinition configDefinition) GetWebhookInterval() time.Duration {
+	// not currently configurable.
+	return time.Second
```
jfberry, line 29: ...and I'm not sure it should be.  I chose a second because small frequent deliveries means less for golbat to hold, and less for receiver to process in each drop.
Since it's all in memory this does not impact the database.
If anything, I'd make it smaller not bigger :-)

comstud: Yes, I agree with all of that. :)

webhooks/sender.go
```diff
@@ -1,52 +1,145 @@
...
-func sendWebhooks(queue WebhookQueue) {
-	req, err := http.NewRequest("POST", queue.url, bytes.NewBuffer(queue.webhook))
+type webhooksSender struct {
+	mutex       sync.Mutex
+	collections webhookCollection
+	webhooks    []*webhook
+	httpClient  *http.Client
+}
+
+func (sender *webhooksSender) getCurrentCollection() webhookCollection {
```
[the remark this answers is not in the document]
comstud: Yes, I'll add that.

webhooks/webhook.go
```diff
@@ -0,0 +1,134 @@
...
+		return fmt.Errorf("failed to send webhook to %s: %s", wh.url, err)
+	}
+
+	defer func() {
+		// full body must be read to reuse keep-alive connections.
+		io.Copy(io.Discard, resp.Body)
+		resp.Body.Close()
+	}()
+
+	// TODO: implement retries upon request fails
```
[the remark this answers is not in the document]
comstud: Yeah, I'm personally just kind of picky and I have a thing for "no downtime upgrades". If something is poorly performing, that is one thing, and I agree with you for that case. Retries are just a big waste of "work" for no reason for that.

What I had in mind was for the times that you are restarting Poracle, for example. I mean, there's probably a very low chance at this, but what happens if in that one second that Poracle is down... you try to webhook a hundo frigibax and it gets dropped? I would be sad. But then.. there's also a low chance that I would ever know, so I don't know. :)

(FWIW, this is a very similar reason to why I added the graceful shutdown and flushing of webhooks in here before exiting.)

I think there's probably a reasonable way to implement it such that if things are backed up, retries get disabled.
And If it were to be done, I'd make it configurable, of course, and defaulting it to off would also be reasonable.

I'll remove this TODO, though. It's not my project so I probably shouldn't be saying that this one is a TODO.

comstud:
> If you can merge up main, I will move this to my production system to test

Yeah, I'll fix those 2 comment-only things and rebase on main in a bit. Thanks for the review.

comstud:
Ready. Rebased on main. Removed the retry TODO and added comment for getCurrentCollection.

webhooks/webhook.go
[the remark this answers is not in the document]
jfberry: If retries were intelligent somehow - eg tried a few times for up to 30 seconds then stopped retrying all attempts at this destination then it could be fine. 
what I would hate is to get into a situation where a down destination was causing a massive spike of retries and held data on golbats side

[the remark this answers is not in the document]
comstud: Agree. That's no good.

jfberry:
I tried to switch to this branch but it appears to be quite out of date w.r.t. main (eg missing migrations etc) so not able to test locally right now

jfberry:
(ignore my previous, operator error! now running on my prod)

Fabio1988 (approved):

PR #159, opened 2023-10-08
Small pokemon api fixes

jfberry:
* Use trunc instead of round for pokemon iv 
* Do not return expired pokemon (v2 api, v1 api untouched)
*


decoder/pokemonRtree.go
```diff
@@ -92,7 +92,7 @@ func updatePokemonLookup(pokemon *Pokemon, changePvp bool, pvpResults map[string
 		Cp:                 int16(valueOrMinus1(pokemon.Cp)),
 		Iv: func() int8 {
 			if pokemon.Iv.Valid {
-				return int8(math.Round(pokemon.Iv.Float64))
+				return int8(math.Floor(pokemon.Iv.Float64))
```
comstud, line 95: Assuming this is %, everyone who ever talks about IV % always rounds. Like, 44/45 IVs is a "98", not a "97". Not sure if that makes a difference here, but just mentioning it.

Fabio1988: I agree, but for endusers this is sometimes not clear :)

I guess with that change api result would differ from db result :/

jfberry: Didn’t mean to resolve 
The db is a float. React map filters these results when sent to it anyway
Since it’s a min/max the round meant that it was returning iv values lower than the minimum

comstud: I think it comes down to... is it a behavior change for people who have been using maps, etc, for a long time? If people are filtering for 98+, I think they would suddenly be surprised to only see hundos if they were seeing the 44/45 ivs before as a 98. I know that I would expect that to be a 98.

comstud: I agree that sometimes it's not clear, though, and I guess to be safe I sometimes find myself filtering for 97+ just to make sure I get those. :)

jfberry: This is an API call, not affecting the results in the DB or the long result returned by the API.
This is to ensure that when you ask for min 98 (an int), you do not get 97.7% pokemon returned.

jfberry: > I agree that sometimes it's not clear, though, and I guess to be safe I sometimes find myself filtering for 97+ just to make sure I get those. :)

Reactmap's implementation of the API when set to 98 was asking for 98, getting the 97.7% returned, and then not displaying it anyway - filtering it out on receipt.

comstud: Yeah, but the API is used for filtering on the map, right? If your slider was at 98 before, does this suddenly change your results to never see 44/45 IVs?

Fabio1988: This was supposed to be like that previously as well :)

jfberry: as I said, the map filtered it out anyway after we sent it. This changes nothing in reactmap, only stops golbat sending mons that are going to be filtered out.
Logically golbat is incorrect because you ask for 'min 98' and get given a line item with a 97.7 in it (and that is actually what it says on the json return because you get the full float in the return)

Fabio1988 (approved):

decoder/pokemonRtree.go
comstud: Race condition on our replies. Got it. :)  If the map filtered it out client side, then it sounds like there's no change... except that you're now properly filtering server side. What do the IV bubbles on map say? 97?

Fabio1988: The bubble rounds up.. so showing 98 instead of 97.7 :)

comstud: That's what I thought. :) Well, that's not consistent. :) Are we sure the correct fix is to not also make ReactMap round, also? Hehe. I totally get that if we only look at the "math" side of this... 97.7 < 98. But I am not sure most users who have played this game for a long time would ever call 44/45 anything but a "98".

I'm just asking questions. :)

jfberry: My view would be that golbat's api should work mathematically and it is for callers to decide whether they ask for 97 to capture all values that would round to 98

comstud: Ok. Good view. +1

PR #161, opened 2023-10-09
Update pyroscope dependency

comstud:
tl;dr: this dependency has moved orgs on github.

Flygon had pyroscope locked at 1.1 and golang 1.20. And pyroscope 1.1 doesn't build for me under golang 1.21. And I found that while 1.2 does build under 1.21 (Golbat uses those versions), it appears that the project has moved orgs on github.

I PRed fixes to Flygon (after discussing with Fabio). And this PR is the corresponding change for Golbat.

### DIFF
--- a/external/pyroscope.go
+++ b/external/pyroscope.go
@@ -1,11 +1,12 @@
 package external
 
 import (
-	"github.com/pyroscope-io/client/pyroscope"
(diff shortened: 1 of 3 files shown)


lenisko:
thanks!

PR #167, opened 2023-10-21
Allow docker build to make use of vendor

comstud:
Having to download dependencies every time during a docker build after a code change makes the build take longer than if dependencies were simply vendored.

This just adds a check to the Dockerfile to see if vendor/ is being used, and if so, skips downloading.

After this lands, one can manage keeping vendor/ up to date in their workspace via 'go mod vendor' and a docker build will not need to download anything. This reduces the docker build time by ~25% for me, but YMMV.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,8 +13,7 @@
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 
-# Dependency directories (remove the comment below to include it)
-# vendor/
+/vendor/
 config.json
 config.toml
 docker-compose.yml
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ FROM golang:1.21-alpine as build
 WORKDIR /go/src/app
 COPY . .
 
-RUN go mod download
+RUN if [ ! -f vendor/modules.txt ]; then go mod download; fi
 RUN CGO_ENABLED=0 go build -tags go_json -o /go/bin/golbat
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -24,4 +23,4 @@ docker-compose.yml
 /golbat
 /geojson/*
 vbase.proto
-__debug_bin
\ No newline at end of file
+__debug_bin
```
comstud, line 26: all that's going on here is that vim just added the missing newline to the end of file here.

Fabio1988:
Hmm that's why GitHub actions are caching the dependencies. I am not sure if that is the same approach or if this affects our github Pipeline.

Pls use the provided images, no need to build on your own... :)

jfberry:
> Pls use the provided images, no need to build on your own... :)

this is a local development scenario I think?

Fabio1988:
> > Pls use the provided images, no need to build on your own... :)
> 
> this is a local development scenario I think?

Exactly, he prefers to build docker image manually :)

comstud:
Right. I am almost never running 'main'. I am usually running the latest of what I've cooked up. :) Gotta test my stuff so I can tell you that my PRs work. ;) In any case, this doesn't change any behavior for anyone who doesn't have a 'vendor'.

If it's a problem, I'll just copy the Dockerfile... no big deal. I PRed it because It felt like it might be beneficial for other devs that may come along and prefer to maintain a 'vendor'.

PR #172, opened 2023-10-26
Refactor prometheus into StatsCollector interface

comstud:
(Redo: This should be fixed, now. When I moved code for Fabio from main, I had a 'statsCollector := ...' and that of course scoped it to main(). There's a statsCollector global that should have been set.)

This adds a StatsCollector interface so that stats implementations can be pluggable. This has the following benefits beyond it being possible for someone to implement something other than prometheus:

* We can now create and default a 'no-op' collector. As things are today, if prometheus is disabled, all of the calls are still doing some amount of work. This PR makes stats collection calls truly no-ops.

* The potential for bugs is reduced. Right now it is very easy to forget a label value, because WithLabelValues() takes a variable number of values. If the number of values does not match the number that were registered, a panic will occur. If we have method signatures with the correct arguments, bugs are limited to the stats collection code, itself.

* Since the stats collector is now pluggable, we can start to test that stats calls are being made when we expect them to be made by creating mock stats collectors in tests and injecting them. This PR does not attempt to add any such tests, yet, however.


main.go
```diff
@@ -3,36 +3,37 @@ package main
...
-	"github.com/go-sql-driver/mysql"
-	_ "github.com/golang-migrate/migrate/v4/database/mysql"
-	_ "github.com/golang-migrate/migrate/v4/source/file"
+	"golbat/stats_collector"
+	"golbat/webhooks"
 )
 
 var db *sqlx.DB
 var dbDetails db2.DbDetails
+var statsCollector stats_collector.StatsCollector
```
comstud, line 36: The last PR was not setting this due to main having ```statsCollector := ```. My bad.

main.go
```diff
@@ -236,7 +237,11 @@ func main() {
 	// Start the web server.
 	gin.SetMode(gin.ReleaseMode)
 	r := gin.New()
-	external.InitPrometheus(r) // init prometheus if enabled
+
+	// choose the statsCollector we will use.
+	statsCollector = stats_collector.GetStatsCollector(cfg, r)
```
comstud, line 242: This source of the crash was this... it was a ```:=``` in previous PR.

Fabio1988 (approved):

comstud:
Ok, I am actually running this now. :)

Fabio1988:
Thanks for fixing previous branch :)

PR #177, opened 2023-11-12
Accurate shiny stats tracking

comstud:
This PR computes shiny stats, but does so by the combination of 'encounterId,accountUsername'. We need to look at every unique encounterId/accountUsername pair, so this is hooked in before we check if an encounter is a duplicate, and only in the encounter paths.

This converts the existing 'pokemonTimings' cache which is keyed by encounterId into a more general encounter cache, adding the map of accounts that have seen the encounter to it.

Shiny stats are not computed per area. After discussion, we determined that area is not very useful and it's too expensive to compute just for these stats. There can be a lot of duplicate encounters, and this avoids the work to compute the areas on duplicates.

For the DB/in-memory stats, they will be updated with area 'world'.

For prometheus, this sends shiny,shundo,snundo stats. For 'any shiny', I'm sending metrics by dex+form Id. And non-shiny counts are submitted as well by dex+form Id. The purpose for this is so that we can compute accurate shiny odds per dex/form.

A metric is also added to count the number of times the same encounterId is seen for both same and different accounts.

### DIFF
--- a/decoder/stats.go
+++ b/decoder/stats.go
@@ -1,15 +1,17 @@
 package decoder
 
 import (
+	"context"
 	"fmt"
 	"os"
+	"strconv"
 	"sync"
 	"time"
 
-	"github.com/jellydator/ttlcache/v3"
 	"github.com/jmoiron/sqlx"
 	log "github.com/sirupsen/logrus"
 
+	"golbat/encounter_cache"
 	"golbat/geo"
 )
 
@@ -27,11 +29,6 @@ type areaStatsCount struct {
 	verifiedReEncSecTotal int64
 }
 
-type pokemonTimings struct {
-	firstWild      int64
-	firstEncounter int64
-}
-
 var pokemonCount = make(map[geo.AreaName]*areaPokemonCountDetail)
 
 // max dex id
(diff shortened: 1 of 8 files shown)


comstud:
I'll add what I have for dashboard or panel when I get a chance.

decoder/stats.go
```diff
@@ -108,18 +104,82 @@ func ReloadGeofenceAndClearStats() {
...
+
+	username := pokemon.Username.ValueOrZero()
+	if username == "" {
+		username = "<NoUsername>"
+	}
+
+	encounterCacheVal := encounterCache.GetOrCreate(pokemon.Id)
+	isNewEncounter := encounterCacheVal.NumAccountsSeen() == 0
+
+	if encounterCacheVal.SetAccountSeen(pokemon.Username.ValueOrZero()) {
```
jfberry, line 128: If I have understood correctly I think this is showing whether a single account makes a duplicate encounter.
I wonder if an interesting check of MITM optimal workflow would be if the encounter has been seen by any account before (the mystery scanner passes this test, lol)

comstud: Yeah. I think it was you who had made a comment on discord that we gain the ability to have a metric on this... So, I added it. The metric for 'a different account has seen it before' is here too. I did them as labels... So you could just add both values for the label to get a total of all duplicate encounters.

comstud: So, it appears this scanner does mostly pass the test. :) Duplicate encounters went away... EXCEPT... on the hour, every hour... there's a lot. What happens on the hour? Is some cache incorrectly purging things every hour?

Screenshot 2023-11-30 at 7 07 54 PM

comstud: That's more a question for the other project. :) But that graph is from the code in this PR.

jfberry: Weather changes

comstud: Yeah, I wasn't thinking. I thought of that, but I'm going to blame lack of sleep on me thinking weather changes were every 15min. :) Thnx

lenisko (approved):
LGTM, haven't checked the logic though.

@jfberry good to merge?

jfberry:
Looks good to me, I'm not sure if this drives any additional updates to the pokemon table (it seems username is not considered as a change, but shiny is -- although this behaviour was already present.  I do remember the code at one stage only 'set' shiny rather than reset shiny

lenisko:
Been running it for a short while as well. A `pyroscope` shows no impact, stats are there. Merging :)

PR #179, opened 2023-11-19
Fix Ditto lookup form

Mygod:
Form for ditto is set to disguise form. For queries, the form should be set to `UNSET`.

This should fix ReactMap filtering Ditto.

### DIFF
--- a/decoder/pokemonRtree.go
+++ b/decoder/pokemonRtree.go
@@ -8,6 +8,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/tidwall/rtree"
 	"golbat/config"
+	"golbat/pogo"
 	"gopkg.in/guregu/null.v4"
 	"math"
 	"strconv"
@@ -81,8 +82,13 @@ func updatePokemonLookup(pokemon *Pokemon, changePvp bool, pvpResults map[string
 	pokemonLookupCacheItem, _ := pokemonLookupCache.Load(pokemonId)
 
 	pokemonLookupCacheItem.PokemonLookup = &PokemonLookup{
-		PokemonId:          pokemon.PokemonId,
-		Form:               int16(pokemon.Form.ValueOrZero()),
+		PokemonId: pokemon.PokemonId,
+		Form: func() int16 {
+			if pokemon.PokemonId == int16(pogo.HoloPokemonId_DITTO) {
+				return 0
+			}
+			return int16(pokemon.Form.ValueOrZero())
+		}(),
 		HasEncounterValues: pokemon.Move1.Valid,
 		Atk:                int8(valueOrMinus1(pokemon.AtkIv)),
 		Def:                int8(valueOrMinus1(pokemon.DefIv)),


Fabio1988:
Do I assume correctly that we store the disguise form only to distinguish the mon when we want to find the right disguise when encountering in game?

Fabio1988 (approved):

Mygod:
Yep.

jfberry:
Why is the form set to a non valid form in the db anyway? Surely this means all webhooks, etc are all wrong?

Mygod:
> Do I assume correctly that we store the disguise form only to distinguish the mon when we want to find the right disguise when encountering in game?

^

Fabio1988:
The form is set for reasons. Users want to know what to search for when they hunt the ditto.
the form could be important when there are two mons with different form on the same location

jfberry:
That feels like the wrong thing to have done and is causing these knock on issues. Because webhooks would also need fixing at least

Fabio1988:
I think that was always the case for ditto, also in RDM :)

jfberry:
I suggest we don’t set the form to the disguise form then this pr is not needed

Mygod:
And? Can't even bother to give any justification?

Mygod:
How about I suggest we remove all Ditto related functionalities from Golbat then this PR is truly not needed.

jfberry:
Because currently we have an invalid Pokemon/form combination and all users of the data are impacted. 
we store disguise Pokemon id elsewhere so would support a disguise form to go along with it
This feels like partial sticking plaster over data stored in the wrong place

Mygod:
(If you do not wish to do that, then this PR is the second best option.)

Fabio1988:
Even RDM is setting form to 0. 

https://github.com/RealDeviceMap/RealDeviceMap/blob/dd7a9fcc4c1b7ed4b040ec2502d9fb2421b5bc16/Sources/RealDeviceMapLib/Modell/Pokemon.swift#L1342

jfberry:
Most applications would assume a valid Pokemon/form combination - I think rdm set form to 0
Fixing just for reactmap is not the best way forward

Mygod:
Setting form to 0 is just introducing another bug to fix down the road. Feel free to choose this obviously worse option.

Fabio1988:
> As I said this fix is not just fix for ReactMap. It fixes for every client that uses this query interface, including RM.

Tell me another one using it 😉 😂

jfberry:
But then in the response we give an invalid form id
And in the webhook 
And in the database
What bug do we introduce by setting it to the correct form for ditto (0)
If storing the disguise form is important then let’s do that.

Mygod:
> Most applications would assume a valid Pokemon/form combination

As you said, Ditto not only does not have a valid Pokemon/form combination, it does not have a valid Pokemon/form/gender/move combination either.

jfberry:
These things are not relevant for identifying the Pokemon on the map whereas you can argue form is. 
but if you believe form is not then we can just set to 0 always like rdm does ?

Mygod:
Just to be clear, you are seriously proposing to replace my "plaster" with an even worse "plaster". If that's the way you like it, then sure.

jfberry:
> Just to be clear, you are seriously proposing to replace my "plaster" with an even worse "plaster". If that's the way you like it, then sure.

good

Mygod:
Ditto is a special case no matter what and my PR does the minimal change, and indeed my PR sets the form to 0 but only for queries. Downstream should expect Ditto to be special no matter what.

Mygod:
Downstream applications should always do a fallback to form 0 for handling missing masterfile updates anyways so it should not break anything.

jfberry:
I think you are arguing that reactmap should be fixed and golbat is ok in its behaviour? Isn’t the point of your pr that downstream clients do not cope with this invalid combination?

Mygod:
It's only for the query interface. The returned results will have the disguise forms as before.

jfberry:
Right and if reactmap is accessing the database the same problem will be shown there?
And similarly any other client searching or processing data

Mygod:
Okay good point.

PR #187, opened 2023-12-04
Prometheus - live stats

lenisko:
Expose live status to prometheus using looped queries.

Exposed metrics:
- `gyms` - current gyms gauge grouped by team_name, in_battle
- `raids` - current raids gauge grouped by level
- `incidents` - current incidents gauge grouped by kind, confirmed
- `lures` - current lures gauge grouped by type
- `quests` - current quests gauge grouped by ar

Disable by default along prometheus integrity, config defaults as below:
```toml
live_stats = false
live_stats_sleep = 120
```

Solves https://github.com/UnownHash/Golbat/issues/178

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -86,9 +86,11 @@ type pyroscope struct {
 }
 
 type Prometheus struct {
-	Enabled    bool      `koanf:"enabled"`
-	Token      string    `koanf:"token"`
-	BucketSize []float64 `koanf:"bucket_size"`
+	Enabled        bool      `koanf:"enabled"`
+	Token          string    `koanf:"token"`
+	BucketSize     []float64 `koanf:"bucket_size"`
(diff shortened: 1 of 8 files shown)


jfberry:
I think 10s is very aggressive for a default refresh rate, considering that every other query that golbat makes has been tuned and re-tuned to be as lightweight as possible.
My view would be that 2 minutes would likely be a more suitable default

jfberry:
It would likely be better to complete the in-memory pokestop and gym access and count based on state transition if this proves to be a problem. Each of these queries would likely require a different and specialised index

lenisko:
@jfberry I might as well toggle a default to off, what do you think?

jfberry:
I think that would make sense. If these vanity metrics are not important enough for someone to activate, then they probably should not be running

PR #188, opened 2023-12-06
Use GetFocuses() and add showcase pokemon form

comstud:
The Focus field has been deprecated in favor of Focuses. Focuses can have multiple types, so this PR adds support for finding the first pokemon in the Focuses array. If Focuses array is empty, I'm having this fall back to Focus for now just in case.

Also: Showcases can feature specific forms. It would be nice to be able to show the restricted form on the map.

This adds showcase_pokemon_form_id to the `pokestop` table and ShowcasePokemonForm to the Pokestop object and the corresponding code to update them, etc.

This requires updated Protos, so a commit from Fabio for them is included.

### DIFF
--- a/decoder/pokestop.go
+++ b/decoder/pokestop.go
@@ -61,6 +61,7 @@ type Pokestop struct {
 	AlternativeQuestExpiry     null.Int    `db:"alternative_quest_expiry" json:"alternative_quest_expiry"`
 	Description                null.String `db:"description" json:"description"`
 	ShowcasePokemon            null.Int    `db:"showcase_pokemon_id" json:"showcase_pokemon_id"`
+	ShowcasePokemonForm        null.Int    `db:"showcase_pokemon_form_id" json:"showcase_pokemon_form_id"`
 	ShowcaseRankingStandard    null.Int    `db:"showcase_ranking_standard" json:"showcase_ranking_standard"`
 	ShowcaseExpiry             null.Int    `db:"showcase_expiry" json:"showcase_expiry"`
 	ShowcaseRankings           null.String `db:"showcase_rankings" json:"showcase_rankings"`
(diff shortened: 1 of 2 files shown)


comstud:
Untested yet. Compiles, though, so it must work, right?

Mygod:
If only you guys would actually slightly consider [what I proposed](https://github.com/UnownHash/Golbat/issues/125#issuecomment-1646939371) smfh.

Fabio1988:
> If only you guys would actually slightly consider [what I proposed](https://github.com/UnownHash/Golbat/issues/125#issuecomment-1646939371) smfh.

we want to prevent json columns... bad for filtering.. but yeah we will see what this feature will bring

Mygod:
>bad for filtering

There is 0 filtering support in ReactMap.

Mygod:
Given that there can only be one showcase every level 14 S2 cell or something, there is probably nothing wrong with filtering JSON to begin with. Is spending an extra 5ms to do JSON filtering a big deal?

Mygod:
>Premature optimization is the root of all evil

comstud:
> @comstud Sorry but I should clarify that none of those were targeted at your PR (they're for @jfberry's initial implementation). Good PR. 👍

All good. Thanks :)

comstud:
For a few days now, I've been running this in combination with: https://github.com/WatWowMap/ReactMap/pull/884

Looked good to me with the 2 sneasel forms in the showcases. As far as I am concerned, this is ready to be merged. However, @TurtIeSocks did note on his corresponding MR that he may make a tweak based on knowing that null/nil form means that any form is allowed. So, perhaps this shouldn't be merged until his PR is final, final. But I just wanted to note that I've been running this w/o issue.

PR #190, opened 2023-12-09
Add DB call count stats

comstud:
I was trying to track down some IO spikes to my DB filesystem. Strangely nothing really turned up in slow query log. So I thought maybe there was a spam of requests happening for some strange reason. So, to check or eliminate golbat, I added these stats. (It turns out this eliminated Golbat and made me look elsewhere for the stupid thing I was doing.)

This might be interesting for others, also, so here it is. And timings might also be useful to add at some point to catch degradation.

### DIFF
--- a/db/dbDetails.go
+++ b/db/dbDetails.go
@@ -1,6 +1,8 @@
 package db
 
 import (
+	"golbat/stats_collector"
+
 	"github.com/jmoiron/sqlx"
 )
 
@@ -8,4 +10,10 @@ type DbDetails struct {
 	PokemonDb       *sqlx.DB
 	UsePokemonCache bool
 	GeneralDb       *sqlx.DB
+}
+
+var statsCollector stats_collector.StatsCollector
+
+func SetStatsCollector(collector stats_collector.StatsCollector) {
+	statsCollector = collector
 }
--- a/db/gym.go
+++ b/db/gym.go
@@ -2,6 +2,7 @@ package db
 
 import (
 	"context"
+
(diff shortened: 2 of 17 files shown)


comstud:
I'm open to different values for the query label. I tried to name them for the function they were performing.

jfberry:
This is quite nice; not sure there is any way to do it but the usual bug that you see is when updates are caused by the same thing too often - eg, update because of a single field flapping. 
This could be done during the update, but we optimised away the generic reflection comparison because it was too slow (it was the hottest part of the code) leading to this being a pain to actually implement as it would be lots of copy/pasta or generated code

comstud:
Was that ever committed to git? I can go back and see what it looked like.

jfberry:
It was, but it was just a call to cmp.Diff rather than our hand rolled comparison. 
there is still a commented out call of the verbose version for debugging so you can watch for flapping but some way (even in a debug mode) of forcing the change field to Prometheus to be able to see them at helicopter view would probably mean these get caught earlier

comstud:
Yeah, I see. hasChanges*() could be re-worked to note the fields that are different vs just returning a bool. But you lose the short-circuiting in the current 'if' statements if you want a full list of fields changing.

Well, I'll keep this in mind. I think that's for another PR, if there's a reasonable solution.

This one is ready to be reviewed/merged as I've been running it for the past 24h.

lenisko:
Hmm... Good idea actually!

As extension, you can also collect metrics directly from DB using prometheus exporters.

PR #191, opened 2023-12-17
feat: api support for geojson geometries

TurtIeSocks:
This extends the existing API endpoints to also support sending a Feature or a Geometry object in the request body. Should maintain 100% backwards compatibility with the added function that normalizes the request body into a Feature. 

- Moves the ApiLocation / ApiGeofence structs into more appropriate places (why are these even separate structs?)
- Converts all pokestop related SQL to use geojson functions instead

### DIFF
--- a/geo/geofence.go
+++ b/geo/geofence.go
@@ -279,19 +279,18 @@ func MatchGeofences(featureCollection *geojson.FeatureCollection, lat, lon float
 }
 
 func (fence *Geofence) toFeature() *geojson.Feature {
-	var ring orb.Ring
-	for _, loc := range fence.Fence {
-		ring = append(ring, orb.Point{loc.Longitude, loc.Latitude})
+	ring := make(orb.Ring, len(fence.Fence))
+	for i, loc := range fence.Fence {
+		ring[i] = orb.Point{loc.Longitude, loc.Latitude}
 	}
-	polygon := orb.Polygon{ring}
 
-	return geojson.NewFeature(polygon)
+	return geojson.NewFeature(orb.Polygon{ring})
 }
 
 func (fence *GeofenceApi) toGeofence() *Geofence {
-	var locations []Location
-	for _, loc := range fence.Fence {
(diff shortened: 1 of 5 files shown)


geo/geofence.go
```diff
@@ -269,3 +277,47 @@ func MatchGeofences(featureCollection *geojson.FeatureCollection, lat, lon float
...
+	polygon := orb.Polygon{ring}
+
+	return geojson.NewFeature(polygon)
+}
+
+func (fence *GeofenceApi) toGeofence() *Geofence {
+	var locations []Location
+	for _, loc := range fence.Fence {
+		locations = append(locations, loc.ToLocation())
+	}
```
comstud, line 295: I've seen this pattern a lot and I have had plans to fix them. This is more efficient as:
```
locations := make([]Location, len(fence.Fence))
for i, loc := range fence.Fence {
	locations[i] = loc.ToLocation()
}
```

When you know the length up front, this is better, as it results in only 1 alloc... instead of possibly many.

geo/geofence.go
```diff
@@ -269,3 +277,47 @@ func MatchGeofences(featureCollection *geojson.FeatureCollection, lat, lon float
 
 	return
 }
+
+func (fence *Geofence) toFeature() *geojson.Feature {
+	var ring orb.Ring
```
comstud, line 282: Similar to the other comment below, you can:
```
ring := make(orb.Ring, len(fence.Fence))
```
and fill in the elements.

jfberry:
It have had it in my mind to move to https://github.com/tidwall/geojson from the orb geojson that we use now. This is much more widely use and has built in fast quadtree/rtree and other high quality geojson utility functions.  Not sure whether it is helpful to you at this stage, but if you find it is easier to use capability here it will save rewriting your changes later.

TurtIeSocks:
This has been running in production without issues for several days now. I tried the other library but unmarshalling seems to be a bit more complex to distinguish between a geometry input and a feature input...

PR #194, opened 2023-12-23
Add stats for invasion_stats, raid_stats & quest_stats

### LINKED ISSUE #144: More Stats!
Currently Golbat only writes Pokemon stats. Add support for other tables that are currently blank and supported for other platforms.

- quest_stats
- raid_stats
- invasion_stats

basically all the stats.

Cronick:
resolves #144 

- Added function to log raids to raid_stats table in DB.
Not sure about the logic for checking new / active raids is fine, but been running for about a day without any issues on my end.

- Added function to log invasions to invasion_stats table in DB.

- Added function to log quest rewards to quest_stats table in DB.
It works, but can probably be done better. So feedback is very welcome :D


jfberry:
Looks good to me; I assume that you are using this in production without any problem?

Cronick:
> Looks good to me; I assume that you are using this in production without any problem?

Previously ran for a day without any issues, but switched to showcase-type yesterday.
Have setup a new instance with this PR now to let it run. Just to be totally sure everything is good.

Fabio1988 (changes requested):
Can you pls update migration file name?

Cronick:
> Can you pls update migration file name?

Renamed to 26 now.

Fabio1988 (approved):

Fabio1988:
the more stats we write now to DB, should we considere a config toggle to enable them? Could be another PR not this, but I guess we should think about it.

Can you pls merge main into this branch, or at least rebase it to be latest? :) @Cronick

decoder/stats.go
```diff
@@ -365,6 +412,156 @@ func updatePokemonStats(old *Pokemon, new *Pokemon, areas []geo.AreaName) {
...
+
+			invasionStats := invasionCount[area]
+			if invasionStats == nil {
+				invasionStats = &areaInvasionCountDetail{}
+				invasionCount[area] = invasionStats
+			}
+
+			// Exclude Kecleon, Showcases and other UNSET characters for invasionStats.
+			if new.Character != 0 {
+				invasionStats.count[new.Character]++
```
Fabio1988, line 491: so this stats is about rocket invasions and we don't care about other incidents, like kecleon, showcases ? :)

decoder/stats.go
```diff
@@ -365,6 +412,156 @@ func updatePokemonStats(old *Pokemon, new *Pokemon, areas []geo.AreaName) {
...
+	}
+
+	locked := false
+
+	// Loop though all areas
+	for i := 0; i < len(areas); i++ {
+		area := areas[i]
+
+		// Check if RaidSpawnTimestamp has changed, then we can assume raid has disappeared or new raid started.
+		if old == nil || old.RaidSpawnTimestamp != new.RaidSpawnTimestamp {
```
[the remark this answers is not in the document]
Cronick: If a new gym is found, then it will run until the new.RaidPokemonId.Int64 != 0 check yeah, and add the matching area to raidCount.

Have tried with truncated gym table multiple times to trigger the flow faster. Havn't seen any panic caused by that, so it should be bare minimal in memory, until it gets a gym with a RaidPokemonId

decoder/stats.go
Cronick: Naming was a bit tricky for when it was Incident related and when invasion started. So could have mixed it up somewhere.

I tried to keep it open at least for adding the posibility to have something like Showcase, Kecleon or future Incidents stats. Just purely focused on invasion for now, which is why Kecleon, Showcases and UNSET is excluded from invasionStats.

decoder/stats.go
```diff
@@ -365,6 +412,156 @@ func updatePokemonStats(old *Pokemon, new *Pokemon, areas []geo.AreaName) {
 	}
 }
 
+func updateRaidStats(old *Gym, new *Gym, areas []geo.AreaName) {
```
Fabio1988, line 415: Can you pls check your DB, unfortunately for raid stats I only get "world" entries, but not proper area entries. For invasion it works quite well :)

Cronick: I had that bug earlier under development and honestly though I had fixed it, as I havn't been able to trigger it again.
Will give it a look and see if I can trigger the behavior. Confused me a bit at the time, and somehow it works on my end right now.

Fabio1988:
One more point... The fences used in Golbat are considered for mon fences.
But they do not contain the fences for quest fences. so this stat are only reflecting the mon fences.

jfberry:
The fences golbat retrieves are purely stats / distribution fences rather than being quests/mons/etc.  They are not "pokemon" or any other type of fences.

Fabio1988:
> The fences golbat retrieves are purely stats / distribution fences rather than being quests/mons/etc. They are not "pokemon" or any other type of fences.

That's true, but the source is different. If I use a mich bigger quest fence than mon fence the quest stat would show world. The fences right now are used to calculate pokemon area stats, used for blissey to process. If I would also add the bigger fence for quests the pokemon stats are wrong

jfberry:
> > The fences golbat retrieves are purely stats / distribution fences rather than being quests/mons/etc. They are not "pokemon" or any other type of fences.
> 
> That's true, but the source is different. If I use a much bigger quest fence than mon fence the quest stat would show world. The fences right now are used to calculate pokemon area stats, used for blissey to process. If I would also add the bigger fence for quests the pokemon stats are wrong

No, they would not be as the pokemon would be in both fences

Fabio1988:
sorry the migration needs another filename update.. to 27 :)

Cronick:
Sorry. Got a bit busy with other stuff lately.
Been testing with this latest commit for a few days. Seems to be better for raid_stats, but hope to really get it tested tommorrow at raid hour, as it makes it a bit easier to spot missing areas.

Fabio1988:
The update stats should be definitely there where it was. You moved it to a place where it updates stats on every seen.

Cronick:
Was also my theory, that the 15 min wait time at worse should not matter much, but somehow raid_stats don't seem happy with it. That it updates on every seen is also what is done for invasions, so it shouldn't give much difference.

Havn't tested this 1:1 but would be funny to do. To see if it behaves differently.

Fabio1988:
> Was also my theory, that the 15 min wait time at worse should not matter much, but somehow raid_stats don't seem happy with it. That it updates on every seen is also what is done for invasions, so it shouldn't give much difference.
> 
> Havn't tested this 1:1 but would be funny to do. To see if it behaves differently.

There is no 15 minute wait.... The 15 minute wait is only there to at leatlst update the updated column more frequently. If there is any change this 15 minute check would not happen

Cronick:
> There is no 15 minute wait.... The 15 minute wait is only there to at leatlst update the updated column more frequently. If there is any change this 15 minute check would not happen

You are right. Misread it, so yeah no reason to change that part. 
Will continue the troubleshooting a bit

decoder/stats.go
```diff
@@ -375,6 +422,158 @@ func updatePokemonStats(old *Pokemon, new *Pokemon, areas []geo.AreaName) {
...
+	locked := false
+
+	// Loop though all areas
+	for i := 0; i < len(areas); i++ {
+		area := areas[i]
+
+		if new.RaidSpawnTimestamp.ValueOrZero() > 0 &&
+			(old == nil || old.RaidLevel != new.RaidLevel ||
+				old.RaidPokemonId != new.RaidPokemonId ||
+				old.RaidSpawnTimestamp != new.RaidSpawnTimestamp) {
```
Fabio1988, line 444: I think we don't need the check for raid level and pokemon id... it can happen that the same legendary raid is spawning two times e.g.. we could simply check for a changed raid spawn timestamp or raid end timestamp :)

Cronick: Hopefully this should be good now in case of Back-to-back raids.

decoder/stats.go
```diff
@@ -375,6 +422,156 @@ func updatePokemonStats(old *Pokemon, new *Pokemon, areas []geo.AreaName) {
...
+				countStats = make(map[int64]areaRaidCountDetail)
+				raidCount[area] = countStats
+			}
+
+			// Check if Raid has started/is active or RaidEndTimestamp has changed (Back-to-back raids)
+			if old == nil || old.RaidPokemonId != new.RaidPokemonId || old.RaidEndTimestamp != new.RaidEndTimestamp {
+				var countRaids = raidCount[area][new.RaidLevel.ValueOrZero()]
+				countRaids.count[new.RaidPokemonId.ValueOrZero()]++
+				raidCount[area][new.RaidLevel.ValueOrZero()] = countRaids
+			}
```
Fabio1988, line 460:
```suggestion
			var countRaids = raidCount[area][new.RaidLevel.ValueOrZero()]
			countRaids.count[new.RaidPokemonId.ValueOrZero()]++
			raidCount[area][new.RaidLevel.ValueOrZero()] = countRaids
```

decoder/stats.go
```diff
@@ -375,6 +422,156 @@ func updatePokemonStats(old *Pokemon, new *Pokemon, areas []geo.AreaName) {
...
+	}
+
+	locked := false
+
+	// Loop though all areas
+	for i := 0; i < len(areas); i++ {
+		area := areas[i]
+
+		// Raid has started/is active.
+		if new.RaidPokemonId.ValueOrZero() > 0 {
```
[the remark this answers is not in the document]
Cronick: Makes sense yeah. Can't remember on top of my head, if there was a specific reason for the split right now.
Will give it a check locally as well just to be sure, but seems good!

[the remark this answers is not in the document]
Cronick: I remember now.

We start with ```if new.RaidPokemonId.ValueOrZero() > 0 {``` to check that the raid has started, before using ```if old == nil || old.RaidPokemonId != new.RaidPokemonId || old.RaidEndTimestamp != new.RaidEndTimestamp {```

As ```old.RaidEndTimestamp != new.RaidEndTimestamp``` will trigger when it's just an egg, and we don't have RaidPokemonId at that point.

Could probably be changed to ```if new.RaidPokemonId.ValueOrZero() > 0 && (old == nil || old.RaidPokemonId != new.RaidPokemonId || old.RaidEndTimestamp != new.RaidEndTimestamp) {``` and then remove the second if statement. 

So we still ensure that we actually have a RaidPokemonId value higher then 0, and don't count on eggs. Will look further into tomorrow.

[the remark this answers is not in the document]
Fabio1988: I checked the code again. if we get the gym again and there is no raid information we reset the raid columns to 0.
I guess we don't need to check for `old` in the first check

Your suggestion makes sense, so you require the RaidPokemonId to be known and it should only count if either the ID or the end timestamp is different.

Fabio1988 (changes requested):

sql/27_stats_area_fence.up.sql
```diff
@@ -0,0 +1,27 @@
+ALTER TABLE `raid_stats`
```
Fabio1988, line 1: I think we can combine the sql query from the other branch to this.. because you already modify a lot.
https://github.com/UnownHash/Golbat/pull/219

decoder/stats.go
Fabio1988, line 442:
```suggestion
		if new.RaidPokemonId.ValueOrZero() > 0 && (old == nil || old.RaidPokemonId != new.RaidPokemonId || old.RaidEndTimestamp != new.RaidEndTimestamp) {
```

that could really make it... guess it's the best check

Cronick: Sorry about the delay. Yeah, have added it now and removed the second if statement.
So hopefully it should be perfect now 😄

sql/27_stats_area_fence.up.sql
Cronick: Done. Also moved area and fence next to date for pokemon_shiny_stats & pokemon_stats, so all tables follow the same format.

Fabio1988 (approved):

decoder/stats.go
```diff
@@ -375,6 +422,154 @@ func updatePokemonStats(old *Pokemon, new *Pokemon, areas []geo.AreaName) {
...
+
+	locked := false
+
+	// Loop though all areas
+	for i := 0; i < len(areas); i++ {
+		area := areas[i]
+
+		// Check if Raid has started/is active or RaidEndTimestamp has changed (Back-to-back raids)
+		if new.RaidPokemonId.ValueOrZero() > 0 &&
+			(old.RaidPokemonId != new.RaidPokemonId || old.RaidEndTimestamp != new.RaidEndTimestamp) {
```
Fabio1988, line 443:
```suggestion
		(old == nil || old.RaidPokemonId != new.RaidPokemonId || old.RaidEndTimestamp != new.RaidEndTimestamp) {
```
you need the old nil check, otherwise this can panic

Cronick: Great catch! 🤦

decoder/stats.go
```diff
@@ -375,6 +422,154 @@ func updatePokemonStats(old *Pokemon, new *Pokemon, areas []geo.AreaName) {
...
+			}
+		}
+	}
+
+	if locked {
+		incidentStatsLock.Unlock()
+	}
+}
+
+func updateQuestStats(pokestop *Pokestop, haveAr bool, areas []geo.AreaName) {
```
Fabio1988, line 509: I think in this is still a logical mistake. a pokestop is updated during day on different events like lure, invasion, powerup_points etc.

In this method you aren't checking for any differences on the quests. so you could count a quest several times during the whole day when events on the stop are happening.

Fabio1988 (changes requested):
quest logic is not acting correctly, sorry :smile:

decoder/stats.go
Fabio1988: hmm maybe my bad.. followed the code where you call it.. you call it in quest update method...... I wonder why I have quest data for today then, as I tested this branch around 9:45

Fabio1988 (approved):

PR #196, opened 2023-12-27
refactor: startup logs

TurtIeSocks:
- Moves the startup log to right after logs are initialized
- Moves sentry and pyroscope to after logs have been initialized
- Fixes the koji log


Fabio1988 (approved):

jfberry:
Lgtm

PR #202, opened 2024-01-11
Add 'total' column to pokemon_shiny_stats

### LINKED ISSUE #201: Issues with Shiny Stats in DB
Most of our stats are de-duped by ID of pokemon, like pokemon_stats and pokemon_iv_stats

But with shiny stats we de-dupe by account+ID, as the same pokemon can be seen by multiple accounts and shiny depends on account

Therefore stats like hundo and shundo can't be compared
Therefore calculating the real shiny odds from DB is impossible, because pokemon_iv_stats total is different from shiny total (because of account+ID de-dupe)

We might need to come up with a valid solution, to make the stats better.

One possible solution might be to add a new column to `pokemon_shiny_stats` to reflect the total... Or we simply stop de-duping by account+ID and we only count the shiny of first account encountering. With the latter the behavior would be identical over all stats.

comstud:
This adds a 'total' column to pokemon_shiny_stats that contains the total number of 'checks'. 'count' remains the number of shinies.

This is being added so that one can use the DB to compute % or odds by comparing 'count' against 'total'.

Fixes #201


decoder/stats.go
```diff
@@ -542,10 +568,36 @@ func logPokemonCount(statsDb *sqlx.DB) {
...
+					"INSERT INTO pokemon_shiny_stats (date, area, fence, pokemon_id, `count`, total)"+
+						" VALUES (:date, :area, :fence, :pokemon_id, :count, :total)"+
+						" ON DUPLICATE KEY UPDATE `count` = `count` + VALUES(`count`), total = total + VALUES(total);",
+					rowsToWrite,
+				)
+				if err != nil {
+					log.Errorf("Error inserting pokemon_shiny_stats: %v", err)
+				}
+			}
+		}
```
comstud, line 601: I was lazy and didn't try to DRY anything here. Going for the quick fix.

comstud:
This is ready for review. I've tested that 'total' is being populated.

decoder/stats.go
```diff
@@ -508,9 +527,16 @@ func logPokemonCount(statsDb *sqlx.DB) {
 					addRows(&nundoRows, pokemonId, count)
 				}
 			}
-			for pokemonId, count := range stats.shiny {
-				if count > 0 {
-					addRows(&shinyRows, pokemonId, count)
+			for pokemonId, checks := range stats.shinyChecks {
+				if checks.total > 0 {
```
Fabio1988, line 531: I guess we should only write if we found a shiny, not if we saw the mon in general, right?

comstud: It has to be written even if no shiny. There could have been a shiny on previous (or later) runs and accurate total is needed.

skelgaard: yes to get an accurate on shiny stats, you need shiny count / by total count, else the data in stats can't be used to calculate a shiny rate seen on that pokemon

PR #203, opened 2024-01-13
Add showcase pokemon type

Cronick:
An attempt from my side to add support for pokemon type on showcases, that we have seen with the Dazzling Dream event.
Will still need support from Reactmap to actually show the info, but now we parse and store the data in Golbat.

Have only tested this briefly so far, but seems to be working.
Can't say if this is a good solution, but trying to learn. So all feedback is appreciated.

### DIFF
--- a/decoder/pokestop.go
+++ b/decoder/pokestop.go
@@ -592,22 +592,22 @@ func (stop *Pokestop) updatePokestopFromGetContestDataOutProto(contest *pogo.Con
 		focussedPokemon = contest.GetFocus().GetPokemon()
 	}
 
-	if focussedPokemon == nil && focussedPokemonType == nil {
+	if focussedPokemon == nil {
 		stop.ShowcasePokemon = null.IntFromPtr(nil)
 		stop.ShowcasePokemonForm = null.IntFromPtr(nil)
+	} else {
+		stop.ShowcasePokemon = null.IntFrom(int64(focussedPokemon.GetPokedexId()))
+		if focussedPokemon.RequireFormToMatch {
+			stop.ShowcasePokemonForm = null.IntFrom(int64(focussedPokemon.GetPokemonDisplay().GetForm()))
+		} else {
+			stop.ShowcasePokemonForm = null.IntFromPtr(nil)
+		}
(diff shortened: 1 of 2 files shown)


TurtIeSocks:
https://github.com/WatWowMap/ReactMap/pull/916 untested but maybe you can try it @Cronick

Cronick:
Together with the ReactMap PR
image

TurtIeSocks (approved):

jfberry:
Looks good to me. Conflicts with #194 for sql migration so will need to be merged seperately

decoder/pokestop.go
```diff
@@ -583,15 +592,21 @@ func (stop *Pokestop) updatePokestopFromGetContestDataOutProto(contest *pogo.Con
 		focussedPokemon = contest.GetFocus().GetPokemon()
 	}
 
-	if focussedPokemon == nil {
+	if focussedPokemon == nil && focussedPokemonType == nil {
```
Fabio1988, line 595: We should probably separate that check, check for pokemon and type separately.
Only one of it might be active and the other might need to be reset, right?

Cronick:
> Looks good to me. Conflicts with #194 for sql migration so will need to be merged seperately

Yeah. Just called both 25, so both could be merged before the other.
Will of course change that.

decoder/pokestop.go
Cronick: I had that in mind, but it seem to default to null (from what I can understand), if the value is not set.
So that's why I didn't add it as a separate check, but can be added as a seperate check, and reset if it's prefered.

Fabio1988: focussedPokemon and focussedPokemonType defaults to null, that's true. But we also need to reset the value on stops.

It might not be an issue with the actual way how showcases work, but imagine next Showcase eould be pokemon again, then the pokemon type would be still stored in db :)

Fabio1988 (approved):

Fabio1988 (approved):

Fabio1988:
Thanks for your contribution :)

PR #205, opened 2024-01-14
Truncate route description

### LINKED ISSUE #204: route descriptions too long to fit into DB column
I'm seeing a number of these:

```
ERRO 2024-01-13 14:26:18 Failed to decode route insert route error: Error 1406 (22001): Data too long for column 'description' at row 1
```

I tried to truncate... which should truncate on character boundary, not octet boundary... but, I've got this now, lol.

```
WARN 2024-01-13 16:53:47 truncating description for route id 'yokohamaWCS3kml'. Orig description: 赤レンガ倉庫の日本最古の荷役用エレベーター前と、桜木町駅前のポケふた(※)「ピカチュウ」を結ぶルート。新港中央広場を通り抜け、ヨコハマエアキャビンを眺めながら汽車道を歩こう。

(WCSとは、ポケモンのゲーム部門・カードゲーム部門・『ポケモン GO』部門・『ポケモンユナイト』部門の4部門の世界一を決める大会のこと。毎年世界中から、各地域や国の激戦を勝ち抜いたプレイヤーが参加し、世界チャンピオンを巡って熱いバトルが繰り広げられます。)

※ポケモンマンホール「ポケふた」とは、ポケモンの絵が描かれたマンホール蓋のこと。1枚1枚がすべてオリジナルデザインで、世界にひとつだけのマンホール蓋です。
ERRO 2024-01-13 16:53:47 Failed to decode route insert route error: Error 1366 (22007): Incorrect string value: '\xE3' for column `golbat`.`route`.`description` at row 1
```

comstud:
I'm seeing a number of these:

```
ERRO 2024-01-13 14:26:18 Failed to decode route insert route error: Error 1406 (22001): Data too long for column 'description' at row 1
```

Rather than increase the VARCHAR(255) at this time, I've chosen to just truncate the description and log info about it. Truncating unicode correctly is a bit of fun.

Fixes #204

### DIFF
--- a/decoder/routes.go
+++ b/decoder/routes.go
@@ -6,9 +6,11 @@ import (
 	"fmt"
 	"golbat/db"
 	"golbat/pogo"
+	"golbat/util"
 	"time"
 
 	"github.com/jellydator/ttlcache/v3"
+	log "github.com/sirupsen/logrus"
 	"gopkg.in/guregu/null.v4"
 )
 
@@ -164,6 +166,15 @@ func saveRouteRecord(db db.DbDetails, route *Route) error {
 func (route *Route) updateFromSharedRouteProto(sharedRouteProto *pogo.SharedRouteProto) {
 	route.Name = sharedRouteProto.GetName()
 	route.Description = sharedRouteProto.GetDescription()
+	// NOTE: Some descriptions have more than 255 runes, which won't fit in our
+	// varchar(255).
+	if truncateStr, truncated := util.TruncateUTF8(route.Description, 255); truncated {
+		log.Warnf("truncating description for route id '%s'. Orig description: %s",
+			route.Id,
+			route.Description,
+		)
+		route.Description = truncateStr
+	}
 	route.DistanceMeters = sharedRouteProto.GetRouteDistanceMeters()
 	route.DurationSeconds = sharedRouteProto.GetRouteDurationSeconds()
 	route.EndFortId = sharedRouteProto.GetEndPoi().GetAnchor().GetFortId()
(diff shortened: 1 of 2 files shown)


comstud:
I've tested this and this fixes the DB fails I was seeing.

decoder/routes.go
```diff
@@ -94,6 +95,32 @@ func saveRouteRecord(db db.DbDetails, route *Route) error {
...
+		defer func(oldDescr string) {
+			route.Description = oldDescr
+		}(route.Description)
+
+		log.Warnf("truncating description for route id '%s'. Orig description: %s",
+			route.Id,
+			route.Description,
+		)
+
+		// this splits on rune boundaries instead of byte boundaries. only
```
jfberry, line 108: r := []rune(text)
   trunc := r[:width]

?

comstud: yeah, I didn't do that because it causes an allocation, supposedly. But I guess when stop to think about it, who cares, in this spot. I'll change it.

comstud:
My len() check was also wrong. This new version has been tested now also.

comstud:
I swear I left a question here to others and myself.... but I don't see it. I probably left it on another PR by accident ha, but:

I'm wondering if this is not the correct spot to do this. I was trying to preserve the original description in here with the defer, but I think maybe this is just kinda dumb.

Maybe Description should be truncated when *Route is initially formed, and just let it be truncated ?

comstud:
Yeah, I think so. I've updated this.

Fabio1988 (approved):

Fabio1988 (approved):

PR #206, opened 2024-01-15
Only gender from encounter is confirmed

Fabio1988:
as there is some rumor at reddit the mon in GMO might show wrong gender, it's different from the encountered/catched one. This bug might be fixed hopefully soon from Nia


jfberry:
I think you need to update wildSignificantUpdate as well at least

Fabio1988:
This was a bug introduced by nia itself, fixed on their side

PR #207, opened 2024-01-16
rework showcase to support buddy focus requirements

Fabio1988:
types:

- pokemon
- type
- alignment
- class
- family
- buddy
- generation
- hatched
- mega
- shiny

Changed the way how we process showcases and those different focuses...

Mid of February 2025 they introduced a buddy min level focus requirement.. This PR will support it

image

image

### DIFF
--- a/decoder/pokestop.go
+++ b/decoder/pokestop.go
@@ -61,6 +61,7 @@ type Pokestop struct {
 	AlternativeQuestTitle      null.String `db:"alternative_quest_title" json:"alternative_quest_title"`
 	AlternativeQuestExpiry     null.Int    `db:"alternative_quest_expiry" json:"alternative_quest_expiry"`
 	Description                null.String `db:"description" json:"description"`
+	ShowcaseFocus              null.String `db:"showcase_focus" json:"showcase_focus"`
 	ShowcasePokemon            null.Int    `db:"showcase_pokemon_id" json:"showcase_pokemon_id"`
 	ShowcasePokemonForm        null.Int    `db:"showcase_pokemon_form_id" json:"showcase_pokemon_form_id"`
 	ShowcasePokemonType        null.Int    `db:"showcase_pokemon_type_id" json:"showcase_pokemon_type_id"`
@@ -184,9 +185,7 @@ func hasChangesPokestop(old *Pokestop, new *Pokestop) bool {
 		!floatAlmostEqual(old.Lat, new.Lat, floatTolerance) ||
 		!floatAlmostEqual(old.Lon, new.Lon, floatTolerance) ||
 		old.ShowcaseRankingStandard != new.ShowcaseRankingStandard ||
-		old.ShowcasePokemon != new.ShowcasePokemon ||
-		old.ShowcasePokemonForm != new.ShowcasePokemonForm ||
-		old.ShowcasePokemonType != new.ShowcasePokemonType ||
+		old.ShowcaseFocus != new.ShowcaseFocus ||
 		old.ShowcaseRankings != new.ShowcaseRankings ||
 		old.ShowcaseExpiry != new.ShowcaseExpiry
 }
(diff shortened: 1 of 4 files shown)


Fabio1988:
useful to refactor showcases... ponyta contained two focusses 
```
METHOD_GET_CONTEST_DATA | 2105

GetContestDataProto {
    fort_id "5c73d33313a03606a6672c26a4e5ef6c.16"
}

GetContestDataOutProto {
    status Status.SUCCESS:1
    
    contest_incident {
        
        contests {
            contest_id "5c73d33313a03606a6672c26a4e5ef6c.16-CUSTOM-MAX-POKEMON_SIZE-1726322400001"
            focus_string_key "pokemon_name_0077"
            
            focus {
                
                pokemon {
                    pokedex_id HoloPokemonId.PONYTA:77
                }
            }
            
            metric {
                pokemon_metric ContestPokemonMetric.POKEMON_SIZE:1
                ranking_standard ContestRankingStandard.MAX:2
            }
            
            schedule {
                
                contest_cycle {
                    start_time_ms 1726322400001
                    end_time_ms 1726444800001
                    contest_occurrence ContestOccurrence.CUSTOM:8
                    custom_cycle_warmup_duration_ms 3600000
                    custom_cycle_cooldown_duration_ms 1
                }
            }
            
            focuses {
                
                pokemon {
                    pokedex_id HoloPokemonId.PONYTA:77
                    require_form_to_match true
                    
                    pokemon_display {}
                }
            }
            
            focuses {
                
                pokemon {
                    pokedex_id HoloPokemonId.PONYTA:77
                    require_form_to_match true
                    
                    pokemon_display {
                        form Form.PONYTA_NORMAL:1011
                    }
                }
            }
        }
    }
}
```

```
METHOD_GET_CONTEST_DATA | 2105

GetContestDataProto {
    fort_id "db3911e7e2fb4487bb5df412cda2a733.16"
}

GetContestDataOutProto {
    status Status.SUCCESS:1
    
    contest_incident {
        
        contests {
            contest_id "db3911e7e2fb4487bb5df412cda2a733.16-CUSTOM-MAX-POKEMON_SIZE-1726322400001"
            focus_string_key "pokemon_name_0077"
            
            focus {
                
                pokemon {
                    pokedex_id HoloPokemonId.PONYTA:77
                }
            }
            
            metric {
                pokemon_metric ContestPokemonMetric.POKEMON_SIZE:1
                ranking_standard ContestRankingStandard.MAX:2
            }
            
            schedule {
                
                contest_cycle {
                    start_time_ms 1726322400001
                    end_time_ms 1726444800001
                    contest_occurrence ContestOccurrence.CUSTOM:8
                    custom_cycle_warmup_duration_ms 3600000
                    custom_cycle_cooldown_duration_ms 1
                }
            }
            
            focuses {
                
                pokemon {
                    pokedex_id HoloPokemonId.PONYTA:77
                    require_form_to_match true
                    
                    pokemon_display {
                        form Form.PONYTA_GALARIAN:2336
                    }
                }
            }
        }
    }
}
```

```
METHOD_GET_CONTEST_DATA | 2105

GetContestDataProto {
    fort_id "cd266ebf35bc4b10be18075898001142.12"
}

GetContestDataOutProto {
    status Status.SUCCESS:1
    
    contest_incident {
        
        contests {
            contest_id "cd266ebf35bc4b10be18075898001142.12-CUSTOM-MAX-POKEMON_SIZE-1726322400001"
            focus_string_key "pokemon_name_0077"
            
            focus {
                
                pokemon {
                    pokedex_id HoloPokemonId.PONYTA:77
                }
            }
            
            metric {
                pokemon_metric ContestPokemonMetric.POKEMON_SIZE:1
                ranking_standard ContestRankingStandard.MAX:2
            }
            
            schedule {
                
                contest_cycle {
                    start_time_ms 1726322400001
                    end_time_ms 1726444800001
                    contest_occurrence ContestOccurrence.CUSTOM:8
                    custom_cycle_warmup_duration_ms 3600000
                    custom_cycle_cooldown_duration_ms 1
                }
            }
            
            focuses {
                
                pokemon {
                    pokedex_id HoloPokemonId.PONYTA:77
                    require_form_to_match true
                    
                    pokemon_display {}
                }
            }
            
            focuses {
                
                pokemon {
                    pokedex_id HoloPokemonId.PONYTA:77
                    require_form_to_match true
                    
                    pokemon_display {
                        form Form.PONYTA_NORMAL:1011
                    }
                }
            }
        }
    }
}


```

Fabio1988:
image

decoder/pokestop_showcase.go
```diff
@@ -0,0 +1,157 @@
...
+		"shiny": p.GetRequireToBeShiny(),
+	}
+}
+
+func createFocusStoreFromContestProto(contest *pogo.ContestProto) map[contestFocusType]contestFocusRequirementProvider {
+	focusStore := make(map[contestFocusType]contestFocusRequirementProvider)
+
+	for _, focus := range contest.GetFocuses() {
+		if pok := focus.GetPokemon(); pok != nil {
+			focusStore[focusPokemon] = &contestPokemonFocus{pok}
```
jfberry, line 126: interfaces are a pointer anyway, can't you just assign this without the address?
I'm also wondering why you don't just create the whole map here rather than  even use interfaces at all? It seems like you return this structure, then immediately call Requirements on it so why not just return an appropriate fully populated map
eg `map[contestFocusType]map[string]any` ?
and then you don't even need any interfaces you just do the extraction in this function so it is simpler to understand.
Perhaps I am missing the purpose of the extra data structures and all the interfaces?

jfberry:
Does this mean that the pokemon columns are not available?
They could be converted to virtual columns extracting from the json for compatibility (we have done that elsewhere for quests).

Fabio1988:
> Does this mean that the pokemon columns are not available? They could be converted to virtual columns extracting from the json for compatibility (we have done that elsewhere for quests).

They are available, I posted a screenshot :)
But yeah can be handled as virtual column as well, then we can get rid of the logic in golbat?! :)

decoder/pokestop_showcase.go
Fabio1988: Hmm you might be right, probably just thought too complicated about it initially, will refactor that bit

jfberry: If you can't have more than one focus, it may as well just return the final map[string]any (but I think you can have more than one focus?)

jfberry:
Ah yes I see you populating it during the loop. It confused me because I probably would have constructed the output, and then pulled it out afterwards to make it clearer what was happening; perhaps it looks clear enough I was only reading it as a diff.

decoder/pokestop_showcase.go
Fabio1988: For now it's just the pokemon focus which appears twice if they use form.... Then there is one focus with pokemon without form and one with the normal form, like it happened for normal/galarian ponyta... Only the galarian ponyta had one focus, with the galarian form... :)

decoder/pokestop.go
```diff
@@ -563,54 +562,56 @@ func (stop *Pokestop) updatePokestopFromGetContestDataOutProto(contest *pogo.Con
...
+		stop.ShowcaseFocus = null.StringFrom(string(jsonBytes))
 
-			if pokType := focus.GetType(); pokType != nil {
-				if focussedPokemonType == nil {
-					focussedPokemonType = focus
-				}
+		// still support old format - probably still required to filter in external tools
+		// could be replaced by a stored column like we do for quests
+		if key == focusPokemon {
+			if pokemonID, ok := focus["pokemon_id"].(int32); ok {
```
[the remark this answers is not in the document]
jfberry: I can't remember if these columns are indexed?  You can create virtual columns that populate from the json but this will likely be slower than just writing them.  You can materialise them permanently but it will also have to parse the json on every update to do so.
If you populate the map in one go, you can just attempt to read the focusPokemon out of the map at the end:
if f, ok := xxx[focusPokemon]; ok {
... 
}
maps are super quick

PR #208, opened 2024-01-17
feat: use `min` & `max` in pokemon search api

TurtIeSocks:
This makes use of the previously ignored `min` and `max` location arguments.

It does a one time haversine comparison between the two points sent in the API request to ensure the distance is shorter than the new config value `tuning.max_pokemon_distance`. 

After the check, it uses the Pythagorean theorem to calculate the box-dist between the two points for comparison with the Pokemon's distance from the center point.


jfberry:
This should have a maximum distance that golbat will accept from client defined in config with reasonable default (like max Pokemon)

PR #217, opened 2024-02-18
Support parsing guarding Pokemon display

### LINKED ISSUE #164: Parse PokemonFortProto.GuardPokemonDisplay
In `PokemonFortProto` there is `GuardPokemonDisplay` which we do not parse at the moment. This causes the frontend to fail to display correct form/costume of the guarding Pokemon.

Related to that, there is also `GuardPokemonLevel` which could be useful?

Mygod:
Fixes #164. Needs testing.


sql/27_guard_display.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE `gym`
+    ADD COLUMN `guarding_pokemon_display` text DEFAULT NULL;
```
Fabio1988, line 2: After guarding_pokemon_id pls

Fabio1988:
Hmm do we really need full display? I guess much better than adding more separate columns 😂

Wonder if this is really sent in the GMO proto...

Mygod:
It is sent because in the overworld you can see these (forms, badges, shiny, etc). It's not separate columns since I don't think anyone would ever want to filter on these data.

decoder/gym.go
```diff
@@ -156,11 +158,32 @@ func calculatePowerUpPoints(fortData *pogo.PokemonFortProto) (null.Int, null.Int
...
 	gym.GuardingPokemonId = null.IntFrom(int64(fortData.GuardPokemonId))
+	display, _ := json.Marshal(pokemonDisplay{
+		Form:          int(fortData.GuardPokemonDisplay.Form),
+		Costume:       int(fortData.GuardPokemonDisplay.Costume),
+		Gender:        int(fortData.GuardPokemonDisplay.Gender),
+		Shiny:         fortData.GuardPokemonDisplay.Shiny,
+		TempEvolution: int(fortData.GuardPokemonDisplay.CurrentTempEvolution),
+		Alignment:     int(fortData.GuardPokemonDisplay.Alignment),
+		Badge:         int(fortData.GuardPokemonDisplay.PokemonBadge),
+		LocationCard:  int(fortData.GuardPokemonDisplay.LocationCard.LocationCard),
```
Fabio1988, line 184: This causes `SIGSEGV: segmentation violation`... as it tries to access `LocationCard` which is not set

Fabio1988 (changes requested):

Fabio1988:
Pls fix the Segmentation violation, would have merged this. tested but it doesn't work

decoder/gym.go
Mygod: Fixed ty.

Fabio1988: Is the fallback to 0 on purpose? What's the value though if available?

Fabio1988: I wonder if we also need to nil check the display in general, empty gyms? Can't test right now

Mygod: Hmm good point. `0` means unset so it is on purpose.

Fabio1988:
@Mygod did you check this with trafficlight or any other proto display tool?
I am running this now for 3 hours, the column is still empty

Fabio1988:
fixed your lazy PR ;) 

```
{"form":946,"costume":0,"gender":1,"shiny":false,"temp_evolution":0,"alignment":0,"badge":0,"location_card":0}
```

Fabio1988 (approved):
seems to work in my testing environment.. so what's the plan on that? will RM support the data?

Mygod:
Yes... RM will support everything. :)

Fabio1988:
> Yes... RM will support everything. :)

without maintainer? :cry:

lenisko:
sounds like @Mygod is willing to maintain it as well!

Mygod:
@TurtIeSocks Please help support this. 🥺

PR #219, opened 2024-03-02
Increase length of area stats name

sabregreen:
Update area and geofence names with varchar(255) to match koji


Fabio1988:
done in #194

PR #220, opened 2024-03-09
Bump go requiremment to 1.22

comstud:
This bumps golang requirement to 1.22.

1.22.1 is out now and while it doesn't appear that we have any workaround for 'for loop variable scope' in Golbat (unless my search failed me)...it would be nice to not have to worry about it going forward.

There's also a new sql.Null generic. And also 'for i := range <integer>' to range from 0 to <integer>-1.

Supposedly memory performance is improved, also.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: '1.21.x'
+          go-version: '1.22.x'
       - name: Install dependencies
         run: go get .
       - name: Test with the Go CLI
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build image
-FROM golang:1.21-alpine as build
+FROM golang:1.22-alpine as build
 
 WORKDIR /go/src/app
 COPY go.mod go.sum ./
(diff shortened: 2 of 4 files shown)


comstud:
Not tested yet.

jfberry:
I don’t think go opts into the language changes without the go mod file version requirement being enforced which then means users have to upgrade too…

comstud:
That's the idea here. This forces users to upgrade so we can continue to develop without worrying about this going forward (if we find the need we need to copy the for loop variable -- we won't need to remember anymore). Also gain some memory performance, MAYBE. But if you're not ready to force users, that's fine. This PR can sit for a bit if you want. I've not even tested it yet.

Fabio1988:
I am fine with enforcing latest golang, as we know it's very easy to update. Although I thought there would be more changes required

jfberry:
I think we will wait for golbat for a bit, fletchling will require 1.22 and we will build dragonite with 1.22

Fabio1988:
hmm now as all tools use go 1.22 maybe it's time to update this as well? :) 
with proper announcement it should be safe, as go can be updated on the fly

jfberry:
Agree, should merge - can announce this at the same time as nests being formally removed (which was indicated to be on 1st July)

Fabio1988:
Sounds great 😃

PR #221, opened 2024-03-09
Add /debug/pprof routes

comstud:
This is copy/pasta from Fletchling which adds the /debug/pprof routes to the gin router. These routes can be used by 'go tool pprof' or curling directly to store profile data.

This will allow us to download profile data to use for 'pgo'.

(https://go.dev/blog/pgo) (https://go.dev/doc/pgo)

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -62,3 +62,4 @@ url = "http://localhost:4201"
 max_pokemon_distance = 100  # Maximum distance in kilometers for searching pokemon
 max_pokemon_results = 3000  # Maximum number of pokemon to return
 extended_timeout = false
+profile_routes = false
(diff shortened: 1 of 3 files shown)


comstud:
Tested and works

jfberry:
No problem with this - just wonder if it should be behind a configuration option because these are unprotected and some users have golbat publically on the internet [I am assuming that the tools don't support adding a custom header to their requests]

comstud:
Good question. The tools don’t necessarily need to talk directly. Can curl and save the data. So adding auth should be fine. If you want config option too I could add at same time.

lenisko:
++ for simple switch with default `false` in config

comstud:
Shoot, I didn't look at this before adding auth. Well, I'll add the switch, too. :)  I think the auth should be there in case someone turns the switch on and has things wide open.

comstud:
Ok, added 'profile_routes' boolean under tuning, defaulting to false.

Fabio1988 (approved):

Fabio1988 (approved):

PR #224, opened 2024-04-09
feat: missing quest api endpoint

TurtIeSocks:
This adds `/api/missing-quests` that accepts a geofence to look up in the Golbat DB which stops are missing either quest set.


Fabio1988 (approved):

jfberry:
This won't be required in its current form so should not be merged as is

jfberry:
...but when quests are fixed in their current state we may come back to finding missing quests and the queries included are worth keeping so please do not delete the branch if you have closed it

PR #225, opened 2024-04-15
Remove nests

jfberry:
Remove nest functionality since this will not be completed in current form


Fabio1988:
We might be able to import fletchling as module and integrate it into Golbat easily, when it's stable :)

jfberry:
That's a seperate question about whether that should be integrated - but this internal counting was not functionally correct and given there are better solutions now no longer adds any value

comstud:
I agree it doesn't make sense for this to continue to live in Golbat in its current form. However, I guess it should be mentioned that this will break nestcollector for those still using or wanting to use it.

There is a path forward for it if people set up Fletchling and the log parsing is adjusted to understand Fletchling logs... as the same information should be available (and more) from Fletchling logs. I'd push people to migrate to Fletchling, but there are a couple things the nestcollector scripts do that are not yet a part of Fletchling. I think those things are also some separate scripts that just look at and twiddle the DB. (One of them is something along the lines of deactivating nests that never have a nesting pokemon). So, I dunno, those may be able to be migrated over as well easily. Not sure.

jfberry:
I don’t think keeping around this inaccurate data so it can be a source of information is a valid reason to keep it - more it should be to create a timeline to migrate and update other tools. 
I would propose that this be merged on eg 1st June

comstud:
> I don’t think keeping around this inaccurate data so it can be a source of information is a valid reason to keep it - more it should be to create a timeline to migrate and update other tools. I would propose that this be merged on eg 1st June

yeah, that's all I felt this needed: a little notice. sounds good to me.

Issue #228, opened 2024-04-16
Stats should populate world everytime


lenisko:
`world` should be populated every time, even when we have correct area. This way we'll get correct global stats.

jfberry:
I kind of agree. At the moment world is unmatched (which probably has a place also) but I chose the name in an unfortunate way. 
this is a breaking change though

lenisko:
we can have one more then `unmatched` that would work as world do now. in normal terms it would be scout outside supported zones

jfberry:
Yes agreed, world should always be added, unmatched should replace current world. would accept pr

Issue #226, opened 2024-04-16
Stats for quests are not filled correctly


lenisko:
I see a weird results that doesn't match what we have in `pokestop` table.

```sql
-- having multiple fences per areas (country, city, province) this should be much greater than pokestop count
SELECT sum(count) FROM `quest_stats` WHERE reward_type = 12 and pokemon_id = 260 and date = CURDATE(); 
-- 141

SELECT count(*) FROM `pokestop` where quest_reward_type = 12 and quest_pokemon_id = 260; 
-- 247
```

Tagging @Cronick

Cronick:
> having multiple fences per areas (country, city, province) this should be much greater than pokestop count

It will only match with a single fence using MatchStatsGeofence, so won't be much higher then pokestop table, but should at least be equal or higher.
https://github.com/UnownHash/Golbat/blob/main/decoder/pokestop.go#L923-L924

On my own instance, it seems to be counting fine. So a bit weird why there is such a difference on yours. Will have a look later and see if I can find something that could be the cause of this.
```
SELECT sum(count) FROM `quest_stats` WHERE reward_type = 12 and pokemon_id = 260 and date = CURDATE(); 
-- 60

SELECT count(*) FROM `pokestop` where quest_reward_type = 12 and quest_pokemon_id = 260; 
-- 38
```

lenisko:
If this helps.. I have from 3 up to 4 fences for each scanned location. Stats on smallest ones are like 1, or don't exist. Top level look better, but not sure if they are reliable.

Fabio1988:
I tried with following query on my setup, no guarantee if query is correct, copied it from [discord](https://discord.com/channels/1083029607919386654/1218743685806821376)

```sql
SELECT SUM(count) as total
FROM (
    SELECT quest_reward_type AS reward_type, COUNT(*) AS count
    FROM golbat.pokestop
    WHERE quest_expiry > UNIX_TIMESTAMP() AND quest_reward_type IS NOT NULL
    GROUP BY quest_reward_type
    UNION ALL
    SELECT alternative_quest_reward_type AS reward_type, COUNT(*) AS count
    FROM golbat.pokestop
    WHERE quest_expiry > UNIX_TIMESTAMP() AND alternative_quest_reward_type IS NOT NULL
    GROUP BY alternative_quest_reward_type
) AS subquery;

# -> 66768

SELECT SUM(`count`) AS total FROM golbat.quest_stats WHERE date = CURDATE()
[...]

PR #229, opened 2024-04-21
Stats changes

lenisko:
Stats changes.

**WARNING**
- Breaking changes (`world` area is applied everytime now, unlike before - only on missing areas)
- Migrations add a new column.

Changes:
- When no areas are found, apply `unmatched` area
- Add `world` area everytime
- Limit stats batches (those errored a loot)
- Add `amount` to quest stats
- Tiny optimizations/changes
- Add mega energy amount to stats

Solves https://github.com/UnownHash/Golbat/issues/228 https://github.com/UnownHash/Golbat/issues/226

### DIFF
--- a/decoder/stats.go
+++ b/decoder/stats.go
@@ -40,6 +40,8 @@ const maxPokemonNo = 1050
 const maxInvasionCharacter = 523
 const maxItemNo = 1614
 
+const batchInsertSize = 200
+
 type shinyChecks struct {
 	shiny int
 	total int
@@ -233,11 +235,16 @@ func updatePokemonStats(old *Pokemon, new *Pokemon, areas []geo.AreaName) {
 	if len(areas) == 0 {
 		areas = []geo.AreaName{
 			{
-				Parent: "world",
-				Name:   "world",
+				Parent: "unmatched",
+				Name:   "unmatched",
 			},
 		}
 	}
+
+	areas = append(areas, geo.AreaName{
+		Parent: "world",
+		Name:   "world",
+	})
 
 	// General stats
 
@@ -426,11 +433,16 @@ func updateRaidStats(old *Gym, new *Gym, areas []geo.AreaName) {
 	if len(areas) == 0 {
 		areas = []geo.AreaName{
 			{
-				Parent: "world",
-				Name:   "world",
+				Parent: "unmatched",
+				Name:   "unmatched",
 			},
 		}
 	}
+
+	areas = append(areas, geo.AreaName{
+		Parent: "world",
+		Name:   "world",
+	})
 
 	locked := false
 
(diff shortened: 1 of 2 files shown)


jfberry:
looks good to me assuming tested

lenisko:
I will know better tomorrow night/morning after quests.
Had no errors today, but N did something nasty, and I can't really call those verified results.

decoder/stats.go
```diff
@@ -63,8 +65,8 @@ type areaInvasionCountDetail struct {
 
 type areaQuestCountDetail struct {
 	count        int
-	pokemonCount [maxPokemonNo + 1]int
-	itemCount    [maxItemNo + 1]int
+	pokemonCount [maxPokemonNo + 1]map[int]int // for each pokemonId[megaCount] keep a count
```
Fabio1988, line 68:
```suggestion
	pokemonDetails [maxPokemonNo + 1]map[int]int // for each pokemonId[megaCount] keep a count
```

Fabio1988: So for mega energy we need to add additional count for pokemonId? isn't that a item?

jfberry: mega energy is per pokemon type though (eg mega energy for bulbasaur).
It's like candy or xl candy for a pokemon type - which are also typed by pokemon and have an amount for the reward

Fabio1988 (approved):

Fabio1988:
so this PR will introduce:
- `world` which contains overall stats
- `unmatched` which contains stats outside of fences
? this means we can't simple sum the `count`anymore, as it's `world = areas + unmatched` ? :)

jfberry:
> so this PR will introduce:
> 
> * `world` which contains overall stats
> * `unmatched` which contains stats outside of fences
> 
> ? this means we can't simple sum the `count`anymore, as it's `world = areas + unmatched` ? :)

Under this PR, the total seen is in 'world' rather than sum of all areas.
So sum of all areas where <> 'world' should be equal to everything else.

Given you could get the total before with a sum, I guess the question is whether this is a beneficial change (I think renaming the old world to unmatched makes things clearer - the question is whether having world as well is a net benefit I think?)

jfberry:
A strong argument for having 'world' in this form is that there can easily be double counting with overlapping areas.  This means that the world entry is everything by itself.  This swings it for me. 
A sum is not going to be accurate unless you take care to have non-overlapping areas.

lenisko:
Hmm, yeah, that's a correct point. I guess I will get rid of it then, as there won't be an easy way to exclude overlapping areas.

jfberry:
> Hmm, yeah, that's a correct point. I guess I will get rid of `world` then, as there won't be an easy way to exclude overlapping areas.

No, I am saying we should have world because otherwise you can't have a correct count

jfberry:
> Wait, right. We get all matched areas under one counts entry. So `world` is doing own job as it's applied for every bag of areas.
Correct. And is the only way to reliably get the total for the whole of golbat

lenisko:
Either way, stats looks much better than before. Got a bit surprised by amount, but we had event change today, hence numbers are doubled.

Sum of my country fences vs world is showing error at less than 0.0002%. Which is correct, as some of my city fences are a bit greater than country. City counts look good as well.

If there are no other questions, ideas - I'm ready to merge.

PR #230, opened 2024-05-09
Properly preserve gym mega evolutions

Mygod:
Finished mega evolutions do not get cleared so we need to preserve when they finish as well.


Mygod:
image
Screenshot of ReactMap for some additional context.

Fabio1988 (approved):

PR #240, opened 2024-11-13
Add total_stationed_gmax to stations

### LINKED ISSUE #239: Add a column indicating gmax pokemon stationed
Since a station might spawn dmax pokemon if a gmax pokemon is stationed, it might be helpful to have this column for filtering. Maybe webhook should be added about this as well.

P.S. I caught a dmax Bulbasaur at a station. It's lv12 with 0 lowest iv.

Mygod:
Fixes #239.

Some more context for justifying the change: https://www.reddit.com/r/TheSilphRoad/comments/1gpavlj/some_observations_on_maxspot_wildspawns/

### DIFF
--- a/decoder/station.go
+++ b/decoder/station.go
@@ -42,6 +42,7 @@ type Station struct {
 	BattlePokemonMove2     null.Int `db:"battle_pokemon_move_2"`
 
 	TotalStationedPokemon null.Int    `db:"total_stationed_pokemon"`
+	TotalStationedGmax    null.Int    `db:"total_stationed_gmax"`
 	StationedPokemon      null.String `db:"stationed_pokemon"`
 }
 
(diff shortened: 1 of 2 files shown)


Fabio1988:
To understand this, the information is already in stationed pokemon, but we want to have an additional column to prevent ugly JSON queries, right? :)

Mygod:
Yes basically. (I would also guess that JSON queries are not as efficient.) Is there a better way to do this? :thinking:

jfberry:
> Yes basically. (I would also guess that JSON queries are not as efficient.) Is there a better way to do this? 🤔

Probably depends on the context records are going to be retrieved.  If it were to be used for filtering stations, for example, then this is probably the way. If it is used as an attribute for a station and the tool (eg ReactMap) has already pulled the entire station record then this is probably superfluous.

Mygod:
Yes. I'm planning to use it for filtering. Do we also need an index for that column?

jfberry:
> Yes. I'm planning to use it for filtering. Do we also need an index for that column?

For reactmap, most often additional indexes don't help as the request will prefer to use the lat, lon index. You should test on large data sets using `EXPLAIN ANALYZE` but my guess would be that this would not be helpful.

Mygod:
Okay. So we can merge this now? 😄

jfberry:
> Okay. So we can merge this now? 😄

Has there been any corroborating information this is useful? There's nothing to stop you building a branch version of ReactMap that uses this -- this is the normal form to experiment with two parallel branches to harden before merging something that is shown to fully work and be useful

PR #244, opened 2024-11-23
Proper support for strong pokemon & Ditto detector+

Mygod:
Closes #242.

* [x] Reimplement Ditto logic in this new framework
* [x] Do strong boosted Pokemon share IVs with strong non-boosted ones? (Guess I'm not finding out unless weather changes mid-hour)


Mygod:
Found an interesting Pokemon reroll: `[POKEMON] An unexpected stat set was seen upon reencountering 786066163 true: (37,13,13,14,0,5) -> (16,11,2,12,0,5)`

Fabio1988:
> Found an interesting strong Pokemon IV reroll: `[POKEMON] An unexpected stat set was seen upon reencountering 786066163 true: (37,13,13,14,0,5) -> (16,11,2,12,0,5)`

After 18:15?

Mygod:
Pokemon v3 in testing now. This new improved Pokemon parser will always have correct IVs for any Ditto! Ditto scenarios to test:

* [ ] >B0
* [ ] >0P
* [ ] 00/0N>0P
* [ ] PN>0P
* [ ] B0>00/[0N]
* [ ] B0>[00]/0N
* [ ] BN>0P
* [ ] BN>0P or B0>[0N]
* [ ] B0>PP/[PN]
* [ ] B0>[PP]/PN
* [ ] B0>BN
* [ ] 0P>00/[0N]
* [ ] 0P>[00]/0N
* [ ] 0P>PN
* [ ] 0P>BN
* [ ] 00/0N>B0
* [ ] 0N>B0 or 0P>[BN]
* [ ] PP/PN>B0
* [ ] BN>B0
* [ ] B0>0P
* [ ] 0P>B0
* [x] B0
* [ ] 0P
* [ ] 0N
* [ ] BN

Mygod:
Every scenario not involving A/B spawn or spawn reroll has been tested.

jfberry:
Thanks for working on this, I need to spend some more time with it to get my head round it.
I don't think it makes sense to unmarshall the internal array each time you see the pokemon, when it could just be an array that is kept on the pokemon object in memory ready to go.  
There's a seperate debate whether it is worth storing in the database (as the only benefit here would be on a restart) - but even if it is then the conversion to/from the binary structure could just be at retrieve / store to disk

Mygod:
@jfberry Thanks for the comments. First, it is not true that the internal struct is unmarshalled every time: it is only done so for (a) encounters (always) or (b) niche scenarios for wild/nearby/map if spawn rerolls or Ditto weather changes. In my quick pprof profiling session, I can't even see any changes relevant in this PR to show up in the diagram. You can see that it is only done so lazily by checking the `internalPopulated` field and the `populateInternal` func.

As for storing in the database, I'd say it's fine. The data stored ends up being only 10-20 bytes long (not accounting for strong spawns since that cannot be tested, but I still expect to be under 30 bytes). Also most big mappers are using in-memory database only anyways.

jfberry:
It's obviously true that multiple unmarshalls will take more cpu and give more work to the garbage collector. I guess I'm saying that I don't see any advantage to doing it fresh each time!  If you weren't writing to DB you wouldn't even have to marshall it ever.

Mygod:
Well the cost seems negligible compared to other things right now. It's not the bottleneck. In my pprof, ~50% of time is used to wait for mariadb to update spawnpoints in `spawnpointUpdateFromWild`.

Not sure about how it affects the garbage collector but performance on my end seems decent enough.

jfberry:
The point is more about something which is clearly unnecessary. I would be happy to receive a PR improving other paths.
we have used Pyroscope in the past to see where time goes in an individual path - I suggest you use that (well certainly want to look at cpu impact overall before merging)

Mygod:
Even with in-memory database only, this saves RAM. Would you rather have a 42-126 bytes overhead on each Pokemon row or 10-20 bytes? 🤔

jfberry:
IO and CPU are the usually constraint that hits golbat - I’d absolutely rather avoid the cpu and GC pressure. But pyroscope will show the proportional impact of the marshalling to the Pokemon path. 
your spawn point profile point is likely because you haven’t yet learnt the spawn points in the area you are processing

Mygod:
image
I guess someone else needs to test this then? 🤔

Mygod:
image
image
image

Stats for encounter decoding.

lenisko:
The amount of data from our staging instance might be too low to show a big difference.

Mygod:
Looking at the graph, it looks like converting pokemon `EncounterId` to string is a big waste of CPU. Why do we store them as string instead of int64 in the database?

Fabio1988:
> Looking at the graph, it looks like converting pokemon `EncounterId` to string is a big waste of CPU. Why do we store them as string instead of int64 in the database? Speaking of unnecessary things to do...

Just because of historic reasons, was introduced like that in RDM times and Golbat was not meant to break compatibility:)

Mygod:
Here's a bit more data. All things considered, proto is really lightweight compared to everything else going on in Golbat... Unmarshal is virtually free, whereas marshal is more expensive but only called during encounter but still not that bad: only 0.02% overhead. CPU is dominated by GC so something needs to be done but probably not this PR.

image
image
image

@jfberry Thoughts?

jfberry:
> Here's a bit more data. All things considered, proto is really lightweight compared to everything else going on in Golbat... Unmarshal is virtually free, whereas marshal is more expensive but only called during encounter but still not that bad: only 0.02% overhead. CPU is dominated by GC so something needs to be done but probably not this PR.

> @jfberry Thoughts?

The GC use would be contributed to by this unnecessary marshalling and unmarshalling.  The extent to which this contributes to the overall pressure is harder to attribute.  My thoughts are that it isn't required unless we're in the pathway writing to the db (as I've also said I'm not convinced of the value of writing it to the DB at all, since restarts are so infrequent outside of development - but I'm seperating that issue in my mind)

Did you run this through a weather or spawn change when a large majority of pokemon every single pokemon changes?  It's really this pathway which is causing a lot of difficulty for golbat on busy systems and the approach in this PR doesn't help that.

So I would suggest as a matter of good hygiene for this PR it should be moved to load/save

Mygod:
Updated PR to turn off writing proto to db by default. No more proto object allocs.
image

Mergeable now? :)

Fabio1988:
Why a config for that "option"? Seems a bit weird, that's nothing someone would switch ever on.. Right? ;)

Mygod:
Just processed a bunch of spawn rerolls. Seems fine.

Mygod (commented):
epic PR merge when?

decoder/main.go
```diff
@@ -382,8 +385,8 @@ func UpdatePokemonBatch(ctx context.Context, db db.DbDetails, scanParameters Sca
 			if err != nil {
 				log.Printf("getOrCreatePokemonRecord: %s", err)
 			} else {
-				pokemon.updateFromNearby(ctx, db, nearby.Data, int64(nearby.Cell), username)
-				savePokemonRecord(ctx, db, pokemon)
+				pokemon.updateFromNearby(ctx, db, nearby.Data, int64(nearby.Cell), weatherLookup, nearby.Timestamp, username)
```
Fabio1988, line 388: are you missing the `/1000` on that param? :)

decoder/main.go
```diff
@@ -399,33 +402,33 @@ func UpdatePokemonBatch(ctx context.Context, db db.DbDetails, scanParameters Sca
 		if err != nil {
 			log.Printf("getOrCreatePokemonRecord: %s", err)
 		} else {
-			pokemon.updateFromMap(ctx, db, mapPokemon.Data, int64(mapPokemon.Cell), username)
+			pokemon.updateFromMap(ctx, db, mapPokemon.Data, int64(mapPokemon.Cell), weatherLookup, mapPokemon.Timestamp, username)
```
Fabio1988, line 405: also here? You do it on line 413, but not here?! :) 

What's the advantage of passing the timestamp at all if it's part of the proto you are passing? :)

decoder/main.go
```diff
@@ -399,33 +402,33 @@ func UpdatePokemonBatch(ctx context.Context, db db.DbDetails, scanParameters Sca
...
+				//log.Infof("Processed stored disk encounter")
 			}
-			savePokemonRecord(ctx, db, pokemon)
+			savePokemonRecordAsAtTime(ctx, db, pokemon, false, mapPokemon.Timestamp/1000)
 		}
 		pokemonMutex.Unlock()
 	}
 }
 
-func UpdateClientWeatherBatch(ctx context.Context, db db.DbDetails, p []RawClientWeatherData) {
```
Fabio1988, line 416: Why did you remove the RawClientWeatherData?! not necessary anymore? normally we collect all proto we need and pass the struct array :)

main.go
```diff
@@ -817,36 +816,32 @@ func decodeGMO(ctx context.Context, protoData *ProtoData, scanParameters decoder
...
 	}
-	if scanParameters.ProcessPokemon {
-		decoder.UpdatePokemonBatch(ctx, dbDetails, scanParameters, newWildPokemon, newNearbyPokemon, newMapPokemon, protoData.Account)
-	}
 	if scanParameters.ProcessWeather {
-		decoder.UpdateClientWeatherBatch(ctx, dbDetails, newClientWeather)
+		decoder.UpdateClientWeatherBatch(ctx, dbDetails, decodedGmo.ClientWeather)
+	}
+	if scanParameters.ProcessPokemon {
+		decoder.UpdatePokemonBatch(ctx, dbDetails, scanParameters, newWildPokemon, newNearbyPokemon, newMapPokemon, decodedGmo.ClientWeather, protoData.Account)
```
Fabio1988, line 844: did you move the pokemon processing after weather because pokemon does need the weather records correctly?

Mygod: Yes.

decoder/main.go
Mygod: Everything is in the proto so no need to create an additional struct. Unnecessary.

decoder/main.go
Mygod: These two actually do take ms since they are expecting `timestampMs` as params.

These timestamps come from remote server so it should be preferred over using local time, which can be out of sync.

Mygod: I messed up but too lazy to change it now. ;)

jfberry:
> ```
> would that help? :) Hopefully I found all occurences, seems to be a valid change
> 
> EDIT: running the changes incl. this patch and it works fine, thanks for this PR :)

I've not been following exactly these timestamp changes - but you need to be careful to include all the precision of the timestamp you get from the GMO. I see you dividing by 1000 very early in this (I haven't had time to follow the flow).
The messages could be being processed out of order from different workers and this should be niantic server time.

Only this time should be used to update spawnpoints - there is a bug where sometimes the encounter time can be used to calculate TTL - this should only happen if there is not a GMO record to work on (I did try and fix this in another PR but lack of testing and motivation stopped me from completing it) - but if timestamp use is being varied in this PR as I see it is, then this should be solved also

decoder/main.go
jfberry: The timestamp in milliseconds from the remote server (only in a GMO) should be the one which is used to determine whether a change is applied (ie latest wins). You can't divide it into seconds as we will be getting lots of GMOs from prospectively multiple workers on the same location if you have raid workers and scanning workers etc

Fabio1988:
I checked the flow... Seems we did that also back then, back in a later state, we never used the timestampMs with full precision it seems... I will have again a proper look to verify...

But I guess using a timestamp as often as possible from protos directly is a good addition :)

jfberry:
> I checked the flow... Seems we did that also back then, back in a later state, we never used the timestampMs with full precision it seems... I will have again a proper look to verify...
> 
> But I guess using a timestamp as often as possible from protos directly is a good addition :)

The intention was (should) be that the timestampMs from the protos is used to determine if this update is more recent than the previous change. This is because GMOs are held (by the go-routine) with the hope that an encounter comes first and then a database update is not required.
If the GMO were processed first and then the encounter arrives later you end up with two DB updates.

This unnescessary with full in-memory processing, and you could argue should not happen in that mode as it will create lots of go-routines that are not needed.

The issue with TTL with encounters first is if the spawnpoint did not have an accurate despawn time yet - we got some pokemon with despawn times an hour out (they are processed just after they despawned because of processing delay) - this is the particular issue I was attempting to fix before.

Fabio1988:
@jfberry yeah can confirm.. the calculculation for the spawnpoints is unchanged in this PR :)

We still use the timestampMs from protos, but we always use `timestampMs/1000` in all occurrences :) so we never use `timestampMs` :) 

so I agree, with this PR and also in the past we would have two DB updates, right? :)

jfberry:
> @jfberry yeah can confirm.. the calculculation for the spawnpoints is unchanged in this PR :)
> 
> We still use the timestampMs from protos, but we always use `timestampMs/1000` in all occurrences :) so we never use `timestampMs` :)
> 
> so I agree, with this PR and also in the past we would have two DB updates, right? :)

I believe we used the timestampMs to decide if it is most recent; that was the intention. It certainly should be the case that for this comparison we do not divide by 1000

Fabio1988:
```diff
diff --git a/decoder/main.go b/decoder/main.go
index c02ad02..d79cffe 100644
--- a/decoder/main.go
+++ b/decoder/main.go
@@ -347,7 +347,7 @@ func UpdatePokemonBatch(ctx context.Context, db db.DbDetails, scanParameters Sca
 				log.Errorf("getOrCreatePokemonRecord: %s", err)
 			} else {
 				updateTime := wild.Timestamp / 1000
-				if pokemon.isNewRecord() || pokemon.wildSignificantUpdate(wild.Data, updateTime) {
+				if pokemon.isNewRecord() || pokemon.wildSignificantUpdate(wild.Data, wild.Timestamp) {
 					go func(wildPokemon *pogo.WildPokemonProto, cellId int64) {
 						time.Sleep(15 * time.Second)
 						pokemonMutex, _ := pokemonStripedMutex.GetLock(encounterId)
@@ -361,7 +361,7 @@ func UpdatePokemonBatch(ctx context.Context, db db.DbDetails, scanParameters Sca
 							log.Errorf("getOrCreatePokemonRecord: %s", err)
 						} else {
 							// Update if there is still a change required & this update is the most recent
-							if pokemon.wildSignificantUpdate(wildPokemon, updateTime) && pokemon.Updated.ValueOrZero() < updateTime {
+							if pokemon.wildSignificantUpdate(wildPokemon, wild.Timestamp) && pokemon.Updated.ValueOrZero() < updateTime {
 								log.Debugf("DELAYED UPDATE: Updating pokemon %s from wild", encounterId)
 
 								pokemon.updateFromWild(ctx, db, wildPokemon, cellId, weatherLookup, updateTime, username)

```

jfberry:
> ```diff
> diff --git a/decoder/main.go b/decoder/main.go
> index c02ad02..d79cffe 100644
> --- a/decoder/main.go
> +++ b/decoder/main.go
> @@ -347,7 +347,7 @@ func UpdatePokemonBatch(ctx context.Context, db db.DbDetails, scanParameters Sca
>  				log.Errorf("getOrCreatePokemonRecord: %s", err)
>  			} else {
>  				updateTime := wild.Timestamp / 1000
> -				if pokemon.isNewRecord() || pokemon.wildSignificantUpdate(wild.Data, updateTime) {
> +				if pokemon.isNewRecord() || pokemon.wildSignificantUpdate(wild.Data, wild.Timestamp) {
>  					go func(wildPokemon *pogo.WildPokemonProto, cellId int64) {
>  						time.Sleep(15 * time.Second)
>  						pokemonMutex, _ := pokemonStripedMutex.GetLock(encounterId)
> @@ -361,7 +361,7 @@ func UpdatePokemonBatch(ctx context.Context, db db.DbDetails, scanParameters Sca
>  							log.Errorf("getOrCreatePokemonRecord: %s", err)
>  						} else {
>  							// Update if there is still a change required & this update is the most recent
> -							if pokemon.wildSignificantUpdate(wildPokemon, updateTime) && pokemon.Updated.ValueOrZero() < updateTime {
> +							if pokemon.wildSignificantUpdate(wildPokemon, wild.Timestamp) && pokemon.Updated.ValueOrZero() < updateTime {
>  								log.Debugf("DELAYED UPDATE: Updating pokemon %s from wild", encounterId)
>  
>  								pokemon.updateFromWild(ctx, db, wildPokemon, cellId, weatherLookup, updateTime, username)
> ```

You may get away with this now with latest go (perhaps not as a pointer) - but the reason timestamp was passed as a function parameter was so that the value was captured

Fabio1988:
true.. my suggested changes are not valid at all.. so I think the state as it is now is way better :) 
My suggested change would break the go-routine behavior, thanks a lot for pointing out :)

decoder/main.go
Mygod: Eventually we should change all s to ms. :)

PR #247, opened 2024-12-09
Use bytebufferpool to reduce gc

Mygod:
`io.ReadAll` used 27.15% of GC alloc. Wow. So wasteful. I wonder if we can get rid of it.

### DIFF
--- a/geo/geofence.go
+++ b/geo/geofence.go
@@ -2,7 +2,7 @@ package geo
 
 import (
 	"fmt"
-	"io"
+	"github.com/valyala/bytebufferpool"
 	"math"
 
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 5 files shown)


Mygod:
I guess this library does not release unused buffers. Maybe there are better ones.

jfberry:
Isn’t this just loaded once?  On a short run this would look disproportionately bad I guess.

Mygod:
Still this might hog up RAM in the long run. I need to find a better library.

jfberry:
Looking at this, it might be on the call every one hour - if it has a leak then there is a real issue (I was having trouble reading through the impact). There is another geojson library we evaluated (if that is where the problem is)

PR #251, opened 2024-12-11
Skip parsing contest if Pokestop is turned off

Mygod:


Fabio1988:
I think the `processed = true` should be moved into the block as well?!  Not sure just noticed it now... @jfberry ?

Mygod:
So merge this then fix that later? :)

PR #259, opened 2025-02-06
Remove pokemonInternalToDb

Mygod:
Improper configuration of this switch causes Pokemon data to be consistent causing Golbat to sometimes crash. This switch is now always on except for PokemonMemoryOnly mode.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -7,24 +7,23 @@ import (
 )
 
 type configDefinition struct {
-	Port                int        `koanf:"port"`
-	GrpcPort            int        `koanf:"grpc_port"`
-	Webhooks            []Webhook  `koanf:"webhooks"`
-	Database            database   `koanf:"database"`
-	Logging             logging    `koanf:"logging"`
-	Sentry              sentry     `koanf:"sentry"`
-	Pyroscope           pyroscope  `koanf:"pyroscope"`
-	Prometheus          Prometheus `koanf:"prometheus"`
-	PokemonMemoryOnly   bool       `koanf:"pokemon_memory_only"`
-	PokemonInternalToDb bool       `koanf:"pokemon_internal_to_db"`
-	TestFortInMemory    bool       `koanf:"test_fort_in_memory"`
-	Cleanup             cleanup    `koanf:"cleanup"`
-	RawBearer           string     `koanf:"raw_bearer"`
-	ApiSecret           string     `koanf:"api_secret"`
-	Pvp                 pvp        `koanf:"pvp"`
-	Koji                koji       `koanf:"koji"`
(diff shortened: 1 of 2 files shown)


jfberry:
A better way to fix this would be to remove this entirely so it is never written / read. 
I don’t support having this always on

Mygod:
It's only on when in-memory only is disabled.

Mygod:
Removing writing breaks stuff and you would need to write many workaround code.

jfberry:
Which means it is writing to the db
And then this option forces is to write more to the db always?

jfberry:
> Removing writing breaks stuff and you would need to write many workaround code.

Surely you just remove the reading entirely and the writing

Fabio1988:
> > Removing writing breaks stuff and you would need to write many workaround code.
> 
> Surely you just remove the reading entirely and the writing

But isn't the scan history used for proper ditto detection?

This was the screenshot of the crash of a user

Screenshot_2025-02-04-15-16-02-48_3481fdfaf807158df2e382fd7e3ae703

jfberry:
> > > Removing writing breaks stuff and you would need to write many workaround code.
> > 
> > 
> > Surely you just remove the reading entirely and the writing
> 
> But isn't the scan history used for proper ditto detection?
> 
> This was the screenshot of the crash of a user

scan history being stored only benefits you with few extra ditto on a restart but costs you increased storage and write time. And the only if your restart crosses an hour boundary. 
so to fix surely better just to remove the ability to read and write scan history and not cost all processing time and storage for this edge case 
In this case a restart would still retain the pokemon but not the scan history 

if there is a bug which means a blank scan history is an issue then this should be solved in combination with this. Just turning on writing this data all the time is costing everyone who does not use memory only efficiency

jfberry:
The original merge was accepted on the basis the default was not to write and incur this cost. There was no evidence gathered on sizing or performance on the basis users would not opt into this in general. 
this reverses this position and means we should not have merged as the pr wasn’t ready.

jfberry:
I have examined this, and surely the fix is just to do the null check for the return on locateAllScans ... it's called in two places only

Mygod:
Yes and no. The issue is what to do when they are null. Since the internal data is missing, you will have to make a guess based on existing data which is a pain. If you want, I can make it simply reset the entire Pokemon struct and treat as a fresh scan.

Again, I don't see why this is necessary. If you want the best quality scan, you should persist internals to db; if you want the best performance, just use in-memory only pokemon which also won't trigger the proto part that has performance concerns from you? This middle ground gives neither and the performance improvement is probably negligible anyway...

Fabio1988:
in favor of #269

PR #260, opened 2025-02-08
Stations webhook

Fabio1988:
Still need to think about how this should look like

With the change of battles after max monday e.g. this could be interesting

Also heard that some users (not admins) don't want to use the map, they just rely on webhooks


JabLuszko:
This needs a lot more of if-s into when creating webhook.
Battle time end can change while station end time does not change (like a 'raid hour') - if mon is the same, nothing will happen.
Bread mode can change while mon_id stays the same - Toxtricity Dyna/Giga Max Raid Day.
Form can change... while rest stays the same - Toxtricity again.
If they introduce Costume mons and just throw them 10 am into even and only costume_id changes?

Basically everything needs to be compared, except moves as those seems to be a placeholder

Fabio1988:
I agree, just a first draft.. will adapt once back from vacation :)
I think it's more about battle details

Fabio1988:
> This needs a lot more of if-s into when creating webhook. Battle time end can change while station end time does not change (like a 'raid hour') - if mon is the same, nothing will happen. Bread mode can change while mon_id stays the same - Toxtricity Dyna/Giga Max Raid Day. Form can change... while rest stays the same - Toxtricity again. If they introduce Costume mons and just throw them 10 am into even and only costume_id changes?
> 
> Basically everything needs to be compared, except moves as those seems to be a placeholder

Thanks again, that summary helped a lot.. almost forgot about toxtricity behavior :D

Mygod:
The current event would be a great opportunity to test this. :eyes:

lenisko (commented):
Haven't run it, but prom code looks ok!

Issue #237, opened 2024-10-08
Save Pokemon form in stats


lenisko:
Expand stats tables with a form of Pokemon.

Currently, we save data in form:
- pokemon_hundo_stats : `date, area, fence, pokemon_id, count`
- pokemon_iv_stats : `date, area, fence, pokemon_id, count`
- pokemon_nundo_stats : `date, area, fence, pokemon_id, count`
- pokemon_shiny_stats : `date, area, fence, pokemon_id, count, total`
- pokemon_stats : `date, area, fence, pokemon_id, count`

lenisko:
solved

PR #266, opened 2025-02-27
Stats: Include Pokemon form (draft)

lenisko:
**DRAFT** https://github.com/UnownHash/Golbat/issues/237

Bugged as I get only single Pokémon & form entry when there's a lot of different Pikachu in there... I will revisit it again someday, unless someone is bored and will find a bug.

```sql
SELECT pokemon_id, form_id, sum(count) FROM `pokemon_stats` group by pokemon_id, form_id ORDER BY pokemon_id, form_id; 
```
```
pokemon_id | form_id | sum(count)
7          | 181     | 6
8          | 184     | 2
10         | 0       | 44
16         | 0       | 48
19         | 45      | 40
20         | 47      | 6
21         | 0       | 28
22         | 0       | 2
25         | 2857    | 228
26         | 49      | 2
27         | 51      | 12
```

### DIFF
--- a/decoder/stats.go
+++ b/decoder/stats.go
@@ -48,11 +48,11 @@ type shinyChecks struct {
 }
 
 type areaPokemonCountDetail struct {
-	hundos      [maxPokemonNo + 1]int
-	nundos      [maxPokemonNo + 1]int
-	shinyChecks [maxPokemonNo + 1]shinyChecks
-	count       [maxPokemonNo + 1]int
-	ivCount     [maxPokemonNo + 1]int
+	hundos      [maxPokemonNo + 1]map[int]int
+	nundos      [maxPokemonNo + 1]map[int]int
+	shinyChecks [maxPokemonNo + 1]map[int]shinyChecks
+	count       [maxPokemonNo + 1]map[int]int
+	ivCount     [maxPokemonNo + 1]map[int]int
 }
 
 type areaRaidCountDetail struct {
@@ -186,8 +186,10 @@ func updateEncounterStats(pokemon *Pokemon) {
 	encounterCache.Put(pokemon.Id, encounterCacheVal, pokemon.remainingDuration(time.Now().Unix()))
 
 	pokemonIdStr := strconv.Itoa(int(pokemon.PokemonId))
+	var formId int
 	var formIdStr string
 	if pokemon.Form.Valid {
+		formId = int(pokemon.Form.ValueOrZero())
 		formIdStr = strconv.Itoa(int(pokemon.Form.ValueOrZero()))
 	}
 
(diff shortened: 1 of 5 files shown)


decoder/stats.go
```diff
@@ -48,11 +48,11 @@ type shinyChecks struct {
 }
 
 type areaPokemonCountDetail struct {
-	hundos      [maxPokemonNo + 1]int
-	nundos      [maxPokemonNo + 1]int
-	shinyChecks [maxPokemonNo + 1]shinyChecks
-	count       [maxPokemonNo + 1]int
-	ivCount     [maxPokemonNo + 1]int
+	hundos      [maxPokemonNo + 1]map[int]int
```
jfberry, line 51: The reason an array was chosen (rather than a map) was because this is an extremely fast data structure and we knew the limits.  If you need a map because this is no longer possible (and I'm sure you're right, form is too sparse to hold in the same way), then you should likely move just to a map of a pokemon type that contains both pokemon id and form id.  There is no longer any advantage to holding the array if you have to go into a map immediately afterwards.

lenisko: I can convert it to array later on once I will confirm it's working as intended. I don't want to rewrite the whole structure in there, it's a mess. As for structure holding Form and Pokemon Id it would require iterating all items.

jfberry: I don’t think you can reasonably convert to an array that’s my point. A single map for pokemon/form is probably the best structure. And of course you can iterate a map by only keys present

decoder/stats.go
```diff
@@ -198,15 +205,26 @@ func updateEncounterStats(pokemon *Pokemon) {
...
+				shinyChecks: make(map[pokemonForm]shinyChecks),
+			}
 			pokemonCount[areaName] = countStats
 		}
-		countStats.shinyChecks[pokemon.PokemonId].total++
+
+		pf := pokemonForm{pokemonId: pokemon.PokemonId, formId: formId}
+
+		entry := countStats.shinyChecks[pf]
+		entry.total++
```
jfberry, line 223: The pattern that was present before is slightly more optimal (ie changing total, and second lookup if shiny) vs the form you landed on with two lookups always. I'm not sure why you changed it?

decoder/stats.go
```diff
@@ -198,15 +205,26 @@ func updateEncounterStats(pokemon *Pokemon) {
 		pokemonStatsLock.Lock()
 		defer pokemonStatsLock.Unlock()
 
-		countStats := pokemonCount[areaName]
-		if countStats == nil {
-			countStats = &areaPokemonCountDetail{}
+		countStats, exists := pokemonCount[areaName]
+		if !exists {
+			countStats = &areaPokemonCountDetail{
+				hundos:      make(map[pokemonForm]int),
```
jfberry, line 211: Do you know how many elements are put in this map? you can pre-size it to save growths (that probably are inevitable)

lenisko: I don't have much data, since that's the first day I run it on prod. But tbh, is there a point of tuning this place that much... ? It was overengineered and ugly as hell to read.

```sql
SELECT COUNT(DISTINCT pokemon_id, form_id) FROM `pokemon_iv_stats` where date = '2025-03-08'
-- 347
```

decoder/stats.go
lenisko: Readability.

jfberry: It's less readable now

decoder/stats.go
jfberry: It's a standard thing to do to correctly size maps (and slices) since they expand and cause GC. It's a couple of characters to do so.

decoder/stats.go
lenisko: Opinion vs opinion. I will change it to get over it :)

decoder/stats.go
lenisko: Well yeah I can add, but would need to gather more data I guess?

jfberry: I was just googling for the boundaries and it appears go 1.24 has switched map implementation so this is likely not needed, so ignore this change

decoder/stats.go
jfberry: You are copying value types needlessly, when this change is made I'll be happy that it at least is no worse than it was (except maps being inherently slower than arrays but this is unavoidable)

lenisko: Can't change it anyway, leaving for future ref:
```go
countStats.shinyChecks[pf].total++
```
> cannot assign to struct field countStats.shinyChecks[pf].total in mapcompiler[UnaddressableFieldAssign](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#UnaddressableFieldAssign)

jfberry (approved):

PR #274, opened 2025-03-28
Update data structures to store pokemon as uint64 in memory

### LINKED ISSUE #248: String conversion for pokemon id & encounters causes excessive cpu
@Mygod reports that he has observed this
[if you have pyroscope screenshots demonstrating it, it would be helpful]

I will play to investigate and mitigate if no-one else looks first

jfberry:
This should result in some memory / cpu / gc advantages - tbc!
Would close #248


lenisko (approved):
Can't see any obvious errors in there, looks good.

Fabio1988 (approved):

Fabio1988 (approved):

PR #275, opened 2025-03-28
spawnpoints - stats & skip aunnecessary check

lenisko:
feat(stats): add spawnpoints to stats
fix(spawnpoint): skip duplication check on new spawanpoint addition


jfberry:
I don’t think this does what you intend. The ‘new’ route is fired when something is not in the cache not if it is a new spawn point. The update mechanism is designed not to have to do a read to discover this. You may be able to use the db result set (not quite sure how) to know if you did an update or an insert

lenisko:
- `newSpawnpoint` is set to `true` and that's only when spawnpoint is missing in cache and db we increment `statsCollector.IncSpawnpoints("new")` 
- `newSpawnpoint` is set to `false` that's when we got ttl.. ah yeah, I need to change this part

lenisko:
I will revisit it later, but "should" work as intended right now

PR #279, opened 2025-04-03
Make pokemon display fields in quest rewards optional

jfberry:


Fabio1988 (approved):

Fabio1988:
tested with freshly scanned area and with latest reactmap, works fine :)

PR #287, opened 2025-05-16
Support collection of rsvp data

jfberry:
Record rsvps against gym table

```json
[{"timeslot":1747387840273,"going_count":1,"maybe_count":0}]
```

An example query to return currently active raids with timeslots in the future:

```sql
select name, from_unixtime(timeslot/1000, '%H:%i:%s') as t, going_count, maybe_count
  from gym, json_table(gym.rsvps,'$[*]' columns (
    timeslot bigint path '$.timeslot', 
    going_count int path '$.going_count', 
    maybe_count int path '$.maybe_count'
    )
  ) as gt 
  where rsvps is not null 
    and gt.timeslot/1000 between raid_spawn_timestamp 
    and raid_end_timestamp and gt.timeslot/1000 > unix_timestamp();
```

(the first check is whether the timeslot is in the currently active raid, the second whether it is in the future - arguably only the second is needed but I have included for completeness of this example)

You may not need to exclude historic data with the latest version of this PR since rsvps are cleared more effectively

### DIFF
--- a/decoder/gym.go
+++ b/decoder/gym.go
@@ -240,7 +240,14 @@ func (gym *Gym) updateGymFromFort(fortData *pogo.PokemonFortProto, cellId uint64
 	if fortData.RaidInfo != nil {
 		gym.RaidEndTimestamp = null.IntFrom(int64(fortData.RaidInfo.RaidEndMs) / 1000)
 		gym.RaidSpawnTimestamp = null.IntFrom(int64(fortData.RaidInfo.RaidSpawnMs) / 1000)
-		gym.RaidBattleTimestamp = null.IntFrom(int64(fortData.RaidInfo.RaidBattleMs) / 1000)
+		raidBattleTimestamp := int64(fortData.RaidInfo.RaidBattleMs)
+
+		if gym.RaidBattleTimestamp.ValueOrZero() != raidBattleTimestamp {
+			// We are reporting a new raid, clear rsvp data
+			gym.Rsvps = null.NewString("", false)
+		}
+		gym.RaidBattleTimestamp = null.IntFrom(raidBattleTimestamp)
+
 		gym.RaidLevel = null.IntFrom(int64(fortData.RaidInfo.RaidLevel))
 		if fortData.RaidInfo.RaidPokemon != nil {
 			gym.RaidPokemonId = null.IntFrom(int64(fortData.RaidInfo.RaidPokemon.PokemonId))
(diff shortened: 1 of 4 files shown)


decoder/gym.go
```diff
@@ -420,7 +456,8 @@ func hasChangesGym(old *Gym, new *Gym) bool {
 		old.PowerUpEndTimestamp != new.PowerUpEndTimestamp ||
 		old.Description != new.Description ||
 		!floatAlmostEqual(old.Lat, new.Lat, floatTolerance) ||
-		!floatAlmostEqual(old.Lon, new.Lon, floatTolerance)
+		!floatAlmostEqual(old.Lon, new.Lon, floatTolerance) ||
+		old.Rsvps != new.Rsvps
```
Fabio1988, line 460: I would move that up at least before floating point comparisons? :)

sql/46_rsvps.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE gym
+    add `rsvps` TEXT default NULL;
```
lenisko, line 2: Why not `JSON` type? Since we put JSON in there, it's safer to use it as we have validation & ability to index virtual columns.

jfberry: Good question.
As background, in mariadb, JSON is an alias for LONGTEXT COLLATE utf8mb4_bin (with a validator check constraint)
All of the other json fields in golbat are defined as text.

Mariadb does not support the -> syntax, you have to use json_extract - which does not require the json type (as it doesn't exist)

lenisko:
Looks good. 

A single point I would argue over is the fact we don't have a quick way to filter out using `going_count` and `maybe_count`. As user might want to ignore any raids with only single interested.

But without making things complicated, I guess that's a correct approach. For now, we could add a small 📖 icon near the raid icon and add a filter to show only those with RSVP data, fetching details inside the raid popup.

jfberry:
> Looks good.
> 
> A single point I would argue over is the fact we don't have a quick way to filter out using `going_count` and `maybe_count`. As user might want to ignore any raids with only single interested.
> 
> But without making things complicated, I guess that's a correct approach. For now, we could add a small 📖 icon near the raid icon and add a filter to show only those with RSVP data, fetching details inside the raid popup.

Obviously you're talking about how something like a map might support this; in general gyms will only be retrieved through the lat/lon indexes as this will be the most efficient, then filters will be applied on these.  You can filter further using json_extract, but being a json array it's never going to be straight forward to write and filtering client side is probably your best bet.  You also need to consider that the timestamp may have passed, and may have been for a previous raid.

jfberry:
> A single point I would argue over is the fact we don't have a quick way to filter out using `going_count` and `maybe_count`. As user might want to ignore any raids with only single interested.

I've included an example sql showing how to use json_table, this is more likely for a static display rather than the way a map would provide a filter

PR #289, opened 2025-05-29
Add API for GetGym

benwilliams95:
Follows existing API framework to allow for Gym data to be accessed via Golbat API in a similar manner to GetPokestop.

The motivating use case of this, can be found in [this PR for Signpost](https://github.com/jfberry/signpost/pull/6). I imagine that there may be other future potential use cases with this as well, since it already exists for Pokemon and Pokestop.

Tested both locally and on-server as part of my main implementation and works as intended.


benwilliams95:
Will take a look

Fabio1988 (approved):

PR #291, opened 2025-06-02
Formid raidstats

HTPaddy:
Added form_id into raidstats table and hopefully updated the primary key to fix the raidlevel if the same pokemon like lapras appears as crypto and normal and not getting counted as the same raidlevel anymore


sql/47_raid_stats_form_id_and_pk_update.up.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE raid_stats
+    ADD COLUMN form_id SMALLINT UNSIGNED NOT NULL DEFAULT 0 AFTER pokemon_id;
+
+ALTER TABLE raid_stats
+    DROP PRIMARY KEY;
+
+ALTER TABLE raid_stats
+    ADD PRIMARY KEY (pokemon_id, date, level, form_id);
```
jfberry, line 8: Previously this was `date`,`area`,`fence`,`pokemon_id` so that there was a natural date order, would recommend that it is now `date`,`area`,`fence`,`pokemon_id`, `form_id`

decoder/stats.go
```diff
@@ -449,50 +451,43 @@ func updatePokemonStats(old *Pokemon, new *Pokemon, areas []geo.AreaName, now in
...
+
+            if raidCount[area] == nil {
+                raidCount[area] = make(map[int64]*areaRaidCountDetail)
+            }
+            countStats := raidCount[area]
+            raidLevel := new.RaidLevel.ValueOrZero()
+            if countStats[raidLevel] == nil {
+    countStats[raidLevel] = &areaRaidCountDetail{count: make(map[pokemonForm]int)}
+}
+pf := pokemonForm{
```
jfberry, line 480: The formatting here is wrong: I am not sure that you are using a go aware editor?
If you execute `go fmt` then it should reformat the file for you

Fabio1988 (approved):

PR #294, opened 2025-07-28
fix: clearing forts on empty cells

Fabio1988:
in case the last fort got removed from S2 Level 15 cell it never got cleared

A user was reporting gyms not being removed, after some deep dive I noticed that it was the last fort within the whole cell. We skipped cleanup in empty cells previously, I removed that check

### DIFF
--- a/decoder/main.go
+++ b/decoder/main.go
@@ -460,17 +460,10 @@ func ClearRemovedForts(ctx context.Context, dbDetails db.DbDetails, mapCells []u
 	now := time.Now().Unix()
 	// check gyms in cell
 	for _, cellId := range mapCells {
-		// lookup for last check
-		if shouldSkipCellCheck(cellId, now) {
-			continue
-		}
-
-		// time to check again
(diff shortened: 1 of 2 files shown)


jfberry:
We skipped the cleanup because we often get empty cells in GMOs when they aren't empty. Does this prevent this situation?

Fabio1988:
> We skipped the cleanup because we often get empty cells in GMOs when they aren't empty. Does this prevent this situation?

No false cleaning in my case yet :)

I guess our solution in general was less error-prune, as we improved a lot after the empty GMO cleanup. I will check the logic more carefully and see if we miss some cases... but in my case there were **49** forts cleaned from my DB. they were just outdated :)

And for now no "Cleared Stop/Gym again in GMO" log which would indicate a bad cleaning :)

But there are for sure cases I did not check yet, like only scan during night / quests... 

I will move that one to DRAFT, as long as I don't know exactly how to proceed :)

PR #297, opened 2025-08-13
Go 1.25 + module update

jfberry:
Update to go1.25 plus module update

A new garbage collector is now available as an experiment. This garbage collector’s design improves the performance of marking and scanning small objects through better locality and CPU scalability. Benchmark result vary, but we expect somewhere between a 10—40% reduction in garbage collection overhead in real-world programs that heavily use the garbage collector.

The new garbage collector may be enabled by setting GOEXPERIMENT=greenteagc at build time. We expect the design to continue to evolve and improve. To that end, we encourage Go developers to try it out and report back their experiences. See the [GitHub issue](https://tip.golang.org/issue/73581) for more details on the design and instructions for sharing feedback.

@Fabio1988 is it possible to update the build to set the GC experiment?

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,7 +4,7 @@ on: [push]
 
 env:
   REGISTRY: ghcr.io
-  GO_VERSION: '1.24.x'
+  GO_VERSION: '1.25.x'
 
 jobs:
   test:


Fabio1988 (approved):
This works, we can try to add the GC flag to the Makefile for testing?

jfberry:
What happens when you just say make? I feel like with two targets you have to do something to specify the default?

Fabio1988:
Yeah with docker it's bit more complicated...  We would need to build two images

jfberry:
I am happy for you to merge it as is, at least there is a way of opting in. For me it feels like green tea is better but with some evidence we could just change the docker build to use it

PR #298, opened 2025-08-13
Experimental move to vtprotobuf

jfberry:


lenisko:
Saw no performance gains from this change. Maybe someone else will confirm as well?

jfberry:
The intention was to reduce gc not improve performance. How did you monitor gc pressure?

PR #300, opened 2025-09-01
Gyms search api - add Geo

lenisko:
Extend Gyms API search with `lat, lon, distance (meteres) || bbox`
Those can be used together with `query` or standalone.

Testing working on our stagging.


lenisko:
Yeah, it grew bigger than I intended.
Reworked and solved all things I guess.
It supports now `name/desc/location-distance/bbox` in multiple combinations with single limit global.

lenisko:
Any comments @jfberry ?

routes.go
```diff
@@ -572,65 +571,101 @@ func GetGyms(c *gin.Context) {
...
+			}
+		}
+		if filter.Bbox != nil {
+			bbox := *filter.Bbox
+			if bbox.MinLat < -90 || bbox.MinLat > 90 || bbox.MaxLat < -90 || bbox.MaxLat > 90 ||
+				bbox.MinLon < -180 || bbox.MinLon > 180 || bbox.MaxLon < -180 || bbox.MaxLon > 180 {
+				c.JSON(http.StatusBadRequest, gin.H{"error": "bbox coordinates out of range: lat must be [-90,90], lon must be [-180,180]"})
+				return
+			}
+			if bbox.MinLat > bbox.MaxLat {
```
Fabio1988, line 640: same applies for Lon?

decoder/api_gym.go
```diff
@@ -0,0 +1,144 @@
...
+		WHERE %s
+		ORDER BY id ASC
+		LIMIT ?
+	`, whereClause)
+
+	args = append(args, search.Limit)
+	q := dbDetails.GeneralDb.Rebind(rawSQL)
+
+	var ids []string
+	if err := dbDetails.GeneralDb.SelectContext(ctx, &ids, q, args...); err != nil {
```
Fabio1988, line 138: I assume that query got that heavy that it doesn't make sense to use the in-memory - cache ?! :)

or wait: you do a SQL query to select gym ids, afterwards you use it to access the in-memory state?

Fabio1988 (approved):

PR #302, opened 2025-09-03
Return the entire pokemon record in /pokemon/id/[id]

ccev:
This aligns the pokemon endpoint with the other /id/ endpoints, where it returns the entire record, not just the coordinates. Backwards compatible & tested.

Previous response:

```json
{
    "lat": 40.758246,
    "lon": -73.980109
}
```

New response:
```json
{
    "id": "16923418122162224990",
    "pokestop_id": "7d4c96c909d447a49f84a65d12bc4948.16",
    "spawn_id": null,
    "lat": 40.758246,
    "lon": -73.980109,
    "weight": null,
    "size": null,
    "height": null,
    "expire_timestamp": 1756909808,
    "updated": 1756908608,
    "pokemon_id": 336,
    "move_1": null,
    "move_2": null,
    "gender": 2,
    "cp": null,
    "atk_iv": null,
    "def_iv": null,
    "sta_iv": null,
    "golbat_internal": null,
    "iv": null,
    "form": 0,
    "level": null,
    "strong": false,
    "weather": 0,
    "costume": 0,
    "first_seen_timestamp": 1756908608,
    "changed": 1756908608,
    "cell_id": -8520149691717911000,
    "expire_timestamp_verified": false,
    "display_pokemon_id": null,
    "is_ditto": false,
    "seen_type": "nearby_stop",
    "shiny": null,
    "username": "iUJDzMhKxno8HEl",
    "capture_1": null,
    "capture_2": null,
    "capture_3": null,
    "pvp": null,
    "is_event": 0
}
```

### DIFF
--- a/decoder/api_pokemon.go
+++ b/decoder/api_pokemon.go
@@ -153,7 +153,7 @@ func SearchPokemon(request ApiPokemonSearch) ([]*Pokemon, error) {
 func GetOnePokemon(pokemonId uint64) *ApiPokemonResult {
 	if item := pokemonCache.Get(pokemonId); item != nil {
 		pokemon := item.Value()
-		apiPokemon := buildApiPokemonResult(pokemon)
+		apiPokemon := buildApiPokemonResult(&pokemon)
 		return &apiPokemon
 	}
 	return nil
--- a/decoder/api_pokemon_scan_v2.go
+++ b/decoder/api_pokemon_scan_v2.go
@@ -202,7 +202,7 @@ func GetPokemonInArea2(retrieveParameters ApiPokemonScan2) []*ApiPokemonResult {
 				continue
 			}
 
-			apiPokemon := buildApiPokemonResult(pokemon)
+			apiPokemon := buildApiPokemonResult(&pokemon)
 
 			results = append(results, &apiPokemon)
 		}


ccev:
@jfberry Could you tell me how? I can't make out a difference in how the data is gathered between SearchPokemon and PokemonOne.

SearchPokemon:

```go
results := make([]*Pokemon, 0, request.Limit)
// ...
if pokemonCacheEntry := pokemonCache.Get(pokemonId); pokemonCacheEntry != nil {
pokemon := pokemonCacheEntry.Value()
results = append(results, &pokemon)
// ...
return results, nil
```

PokemonOne:

```go
if item := pokemonCache.Get(pokemonId); item != nil {
    pokemon := item.Value()
    return &pokemon
}
```

jfberry:
The json tags on the pokemon object are effectively deprecated (should really be removed) since some items need to be recalculated. the search takes them and translate them into an api pokemon object that contains the right details 

not in front of code at the moment.

ccev:
Thanks for clarifying. I can't see where this would be done in PokemonSearch. I'd appreciate if you could take a closer look when you get the time. Thank you!

ccev:
I think I was able to implement the suggestion. Feel free to take another look

decoder/api_pokemon_common.go
```diff
@@ -0,0 +1,61 @@
+package decoder
+
+import (
+    "github.com/UnownHash/gohbem"
+)
+
+func buildApiPokemonResult(pokemon Pokemon) ApiPokemonResult {
```
jfberry, line 7: Just take a pointer here (ie `pokemon *Pokemon`) and it looks perfect. It could return a pointer also I guess since it's using it as a pointer (and returning the pointer gets deserialised also)

ccev: Should be good now 🙂

PR #304, opened 2025-09-12
Parse SharedRouteProto.ShortCode

### LINKED ISSUE #303: CreateRouteShortcodeProto?
Currently invisible routes can be activated by entering a "route share code". However, this code is not present in `SharedRouteProto.ShortCode`. Rather, the client needs to proactively request `CreateRouteShortcodeProto` to obtain the short code. Any chance we can support this?

Mygod:
Fixes #303.

Testing as soon as I get data. :)

Is there anywhere in the entire stack that cleans up route? Ideally we should probably keep the routes with short codes unless it is known to have been deleted or something.

### DIFF
--- a/decoder/routes.go
+++ b/decoder/routes.go
@@ -132,6 +132,7 @@ func saveRouteRecord(db db.DbDetails, route *Route) error {
 			`
 			UPDATE route SET
 				name = :name,
+				shortcode = :shortcode,
 				description = :description,
 				distance_meters = :distance_meters,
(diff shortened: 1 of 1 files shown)


Fabio1988 (approved):

Fabio1988:
No we don't clean them yet... initially there were not that many, but now the whole map view is polluted with routes and nobody can really find a perfect one on a certain path

Mygod:
Sounds good. I see that ReactMap does filter routes by recently seen.

PR #306, opened 2025-09-25
Clear gym defenders when team assignment changes

Mygod:
ReactMap currently displays outdated defender information after gym changes. Until new defender information arrives, this will make RM display the information available in GMO.


jfberry:
I'm not sure this is a valid change because you can't guarantee which order these will be processed
I would suggest a better solution is reactmap not showing the defenders when they don't line up.

jfberry:
Yes, gym defender scans happen when drago detects that a gym LastModifiedMs (in the proto) has changed, up to a certain limit (50 per gmo process, which it doesn't normally hit). 
The GMO update woud normally beat the gym defender update for sure.

I think you're right that this is relatively safe if GymInfoOutProto updates the team colour also.

PR #308, opened 2025-09-26
Fix #116

### LINKED ISSUE #116: Pokemon Webhook might miss pvp value on weather change and weatherboost stays
As we preserve IV values now on weather change when mon is still boosted we don't send pvp value in webhook anymore.

This is happening due to the pokemon processing rework / ditto detection improvement of @Mygod. As we don't store pvp value in Cache (to reduce memory usage) we should either recreate pvp value on the fly or find another way :)

Mygod:
Fixes #116.


Mygod:
Okay redone the fix as you suggested. $0.33 please. :)

```To continue this session, run codex resume 01998875-0efc-70e0-b9ca-fb05bc76f504.```

jfberry (approved):

PR #311, opened 2025-10-11
feat: invasion_lineup_stats

Mygod:
Bring back malte.im to your ReactMap?

(I couldn't test this because I don't have lineup scanning data sent to me.)

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -15,7 +15,7 @@ incidents = true        # Remove incidents after expiry
 quests = true           # Remove quests after expiry
 tappables = true        # Remove tappables after expiry
 stats = true            # Enable/Disable stats history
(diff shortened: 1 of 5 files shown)


Mygod:
It's reverted (for the second time) for now. :)

jfberry:
> It's reverted (for the second time) for now. :)

Appols - the three lineups together in one row I think is the right structure. It was the addition of reward that I was incorrect about - as I’ve had this confirmed that it’s not constant. 
however my point about whether we can still get this data easily going forward stands — the battle has moved server side and requires web socket comms. It’s certainly not from the current Protoss (assuming niantic do not roll back!)

Mygod:
This PR does not contain reward pokemon. It appears to me that Golbat does not currently even process the battle protos beyond the first step, and I still don't have data so I can't tell. :)

PR #315, opened 2025-10-29
Defender fixes

### LINKED ISSUE #313: Gym defender parsing cannot be turned off
It should be turned off when gym parsing is off?

Mygod:
Fixes #313. Can't believe we are throwing away all these data.


jfberry:
Superceded by #363

PR #353, opened 2026-03-31
Persist and expose multi-battle station state

Mygod:
## Why

Max Battles can temporarily override a station's normal local battle. Golbat previously stored only one flat `battle_*` payload on `station`, so a temporary event battle would overwrite the local one and the original state would be lost for planning and downstream filtering.

This PR adds first-class multi-battle support for stations while keeping the existing flat station/webhook fields for compatibility.

## Ground rule

Latest observed game data is treated as ground truth.

For a given station:
- newly observed battles are persisted independently
- if a new battle has a later or equal `battle_end` than another known battle on the same station, the older/equal-ending row is treated as obsolete and removed
- a shorter battle does **not** remove a longer known battle

This keeps temporary overrides and later event rotations aligned with the most recent game state instead of waiting for old rows to expire naturally.

## What this PR adds

- a new `station_battle` table via migration `54_station_battle.up.sql`
- station battle persistence separate from the parent `station` row
- a compatibility projection that still fills `station.battle_*`
- a `battles` array in station API and `max_battle` webhook payloads
- multi-battle-aware station filtering instead of single-battle-only filtering
- hydration of `station_battle` rows in both preload and lazy station reads
- expired `station_battle` cleanup support
- canonical-battle-based metric dedup

## Canonical battle behavior

A station can now know about multiple non-expired battles, but one battle is still chosen as the canonical representative for:
- flat `station.battle_*`
- top-level `max_battle` webhook fields
- battle metric dedup

Selection rule:
- prefer an active battle over an upcoming one
- if none are active, use the best upcoming battle
- later `battle_end` wins among surviving rows

The full `battles` array remains available for planning use cases.

## Identity / persistence

`bread_battle_seed` is used as the persisted battle identity.

Live validation before implementation showed:
- no observed seed reuse across multiple stations
- seeds changed when battles rotated

That made seed the best available battle-instance identifier for persistence.

## API / webhook impact

Backward compatibility is preserved:
- existing top-level `battle_*` station fields remain
- existing top-level `max_battle` webhook fields remain

New additive fields:
- station API: `battles`
- `max_battle` webhook: `battles`

This allows downstream consumers to keep working unchanged while giving planning-oriented consumers the full set.

## Deployment notes

- migration `54_station_battle.up.sql` is included and runs automatically on startup
- no manual backfill from legacy `station.battle_*` into `station_battle` is included
- population of `station_battle` depends on normal rescans after deploy
- optional cleanup support is available via `cleanup.station_battles`

Practical rollout expectation:
- after deploy, rescans repopulate the new table quickly
- during the current non-overlap period, most stations will still have exactly one battle row

## Validation

Tested with:
- `go test ./decoder -run 'StationBattle|StationWebhooks|BuildStationResult|FortLookup|SyncStation|Preload'`
- `go test ./... -run '^$'`

Post-deploy live DB checks on one environment showed:
- migration applied cleanly at version `54`
- no active legacy-only stations
- no active projection mismatches between `station` and `station_battle`
- no active `station_battle` rows missing boss data
- currently one active battle row per station, which matches the present daily-rotation state

## Notes

This PR is intended to preserve battles for planning, especially around Monday / Max Battle event overlaps, while keeping existing API and webhook consumers stable.

Fixes #352.


Mygod:
ReactMap PR is done. From my testing, everything seems ok. Feel free to complain now. :)

I'm also planning to do another review pass in two days.

jfberry:
I haven't done a manual review of it yet, but here is the claude first pass review - which picks up some issues (mainly around the database write not being batched like all the other writers - without looking it's likely to be a msunderstanding of the way sqlx can construct a statement from an array).

PR #353 Review — Persist multi-battle station state

  The PR has clearly been revised since my earlier comments — forceSave/skipWebhook, the synchronous-in-hot-path writes, and the unprotected
  xsync.Store race in GetStationRecordReadOnly are all gone. It's much closer to project conventions now. A stationBattleQueue TypedQueue[string,
   StationBattleWrite] exists; hydration runs under the station lock; the rtree update is fed from the same snapshot used by save/webhook.

  Record access semantics

  - Hydration in both branches of GetStationRecordReadOnly now happens after station.Lock(caller), so the previous race on
  stationBattleCache.Store is gone.
  - getOrCreateStationRecord only hydrates when loading a fresh record from DB. A station that was created via proto-only path (newRecord=false
  after the initial save) never has its battles hydrated until somebody calls GetStationRecordReadOnly, but hasHydratedStationBattles is set the
  first time a proto carries battles via syncStationBattlesFromProto, so this is fine in practice.
  - Cache-hit hydration error in GetStationRecordReadOnly is only Debug-logged and still returns the station (station_state.go ~L83). The DB-load
   branch returns the error. Inconsistent — and a transient SELECT failure leaves the station permanently un-hydrated for that session because
  nothing flips hasHydratedStationBattles. Recommend marking hydrated even on empty/failure or returning the error. Minor.

  Deadlocks / lock ordering

  - No new cross-entity lock acquisitions. All battle state is keyed by stationId and accessed under the station mutex (or via lock-free
  xsync.MapOf for the read paths). Consistent with CLAUDE.md "Lock Ordering" rules.
  - hydrateStationBattlesForStation does a synchronous DB SELECT while holding the station lock. That's the same pattern loadStationFromDatabase
  uses, so it's not a regression, but it does mean a slow station_battle query stalls every other goroutine waiting on that station. Fine in
  steady state because hydration only runs once per station per process lifetime.

  Write-behind correctness

  - stationBattleQueue is a real TypedQueue keyed by StationId, so repeated updates within a flush window coalesce — good.
  - However, flushStationBattleBatch just loops the snapshots and calls storeStationBattleSnapshot one-at-a-time. Each call does BEGIN →
  bulk-INSERT … ON DUPLICATE KEY UPDATE → DELETE … WHERE seed NOT IN (…) → COMMIT. So a flush of 50 stations = 50 transactions × 4 round-trips =
  ~200 round-trips against GeneralDb. The BatchSize knob doesn't actually batch across stations the way other entities' flushes do. Not broken,
  but the queue is providing coalescing only, not true batching.
  - The DELETE … NOT IN (…) strategy is correct for pruning stale seeds without losing newly-inserted rows in the same tx, and the per-snapshot
  tx scope means a flush failure for one station doesn't poison others. Good.
  - Coalescing risk: when a snapshot is enqueued, the Battles slice is cloned (stationBattleWriteFromSlice → cloneStationBattles), so the queue
  holds a stable snapshot even if the in-memory cache mutates afterwards. Good.

  Database write volume

  - For active stations the write trigger is battleListChanged (signature mismatch). The signature is computed from the non-expired battle list,
  so the signature changes whenever:
    a. a new battle arrives (expected),
    b. a known battle expires across now boundary (NOT a real change in DB-relevant data, but still triggers an enqueue).
  - Effect: every station with an active battle will write at least once when that battle ends, even if no proto for it arrives. Bounded — one
  extra write per battle expiry. Acceptable.
  - The 15-min debounce is still bypassed by battleListChanged, so high-churn stations can write more often than before. Should be fine given the
   queue coalesces, but worth knowing.

  Memory

  - Per-station: a []StationBattleData (typically 1–3 entries, ~200 B each) in stationBattleCache + an empty struct entry in
  stationBattleHydratedCache. For ~100k stations, well under 100 MB worst-case. No leak — clearStationBattleCaches is wired into the station
  cache eviction callback. Good.
  - The eviction callback is now installed unconditionally (not gated on FortInMemory), which is necessary for battle-cache cleanup and
  incidentally fixes a small consistency issue.

  CLAUDE.md / pattern adherence

  - ✅ Write-behind queue used.
  - ✅ Cache eviction callback drops auxiliary state.
  - ✅ Hydration done under entity lock.
  - ✅ Webhook payload built from a single snapshot, not by re-reading mutable state.
  - ⚠️ StationBattleData still doesn't follow the standard "embedded Data + dirty + oldValues" entity pattern — but it's a child collection of an
   existing entity, not a standalone entity, so the deviation is justified. The BattleListSignature in StationOldValues is the dirty-tracking
  equivalent.
  - ⚠️ stationBattleSignatureFromSlice is awkward — fields are written in a strange order with Valid flags interleaved between value writes
  (station_battle.go ~L865). It works, but a bytes.Buffer/fnv hash or struct equality would be cheaper and less error-prone. Nit.
  - ⚠️ Webhook regression: createStationWebhooksWithSnapshot returns early if currentSignature == "". That suppresses a webhook in the "last
  battle just ended" transition. Previously a station whose BattleEnd rolled past now would still get a MaxBattle webhook on the EDIT. May or may
   not be intentional — worth confirming with the author.

  Bugs / smells

  1. Cache-hit hydration error swallowed (above) — minor.
  2. fortRtreeUpdateStationOnGet no longer called inside GetOrSetFunc — moved outside. If two goroutines race, only the winner's loadedFromDb is
  true; the loser sees loadedFromDb=false and won't update the rtree. The winner does update it. Correct, though subtle.
  3. StartStationBattleExpiry cleanup interval is 1h13m (stats.go). The other expiry tickers use similarly arbitrary primes — consistent with
  existing convention.
  4. Signature-driven save bypasses 15-min debounce — see above. Worth a one-line comment in the code.
  5. stationBattleSignatureFromSlice ordering — see above. Nit, not a bug.

  Recommendation

  1. Confirms the "battle ended → no battles left" webhook suppression is intentional (or restores it).
  2. Either logs the cache-hit hydration error at Warn/Error or treats it like the DB-load path.

Mygod:
I identified some overengineering in this branch. Once I fix that I will take a closer look at these.

Mygod:
Codex:

> Addressed the batching concern on the current head.
> 
> station_battle now flushes like the other typed write-behind queues instead of only coalescing by station_id: flushStationBattleBatch does one bulk upsert across the flushed rows and one bulk prune across the flushed station ids in a single transaction. I intentionally kept the failure semantics aligned with the existing TypedQueue behavior used elsewhere in Golbat, so batch-level failure isolation is the same as the other batched writers rather than preserving a station_battle-specific special case.
> 
> The earlier notes about the old hydrated-cache/canonical helper paths are stale after the recent cleanup and no longer apply to the current code.
> 
> I did not take the separate seed-collision/schema concern in this branch.

I believe the battle seed is unique (it appears to be a random int64) so defensive primary key by station id is most likely unneeded.

jfberry:
Claude is generally a fan (with some minor comments), I'll probably have some time to look myself in a few days time.


⏺ PR #353 — Re-review of the latest revision (f8e40cd)

  This revision is a substantial cleanup. Most of my prior concerns are resolved.

  What got better

  1. Real batched flush. flushStationBattleBatch is now a single transaction per flush: one bulk INSERT … ON DUPLICATE KEY UPDATE across every
  station in the batch, one DELETE … LEFT JOIN keep_rows to prune obsolete seeds, one commit. The previous "loop and call
  storeStationBattleSnapshot per station" pattern is gone, and with it the concern that the queue was providing only coalescing and no batching.
  DB round-trip volume now scales with BatchSize like the other entities.
  2. Single cache instead of two. stationBattleCache is now xsync.MapOf[string, stationBattleState] carrying {Battles, Loaded}. The parallel
  stationBattleHydratedCache is gone, eliminating the small window where the two maps could disagree. storeStationBattles is the one entry point.
  3. No more "canonical vs projection" duality. stationBattleFromStationProjection, canonicalStationBattleFromSlice,
  stationBattleMatchesProjection, canonicalBattleSeed, restoreStationBattleProjectionFromOldValues — all deleted. The "top" battle is now just
  snapshot.Battles[0] after sortStationBattlesByEnd. StationOldValues shrinks to {EndTime, HasTopBattle, TopBattleSeed, BattleListSignature}.
  Much easier to reason about.
  4. Single source of truth for battle fields on Station. updateFromStationProto no longer pokes battle fields directly — they flow exclusively
  through syncStationBattlesFromProto → applyStationBattleProjection. The previous risk of the proto path and the cache path disagreeing is gone.
  5. Tests consolidated into a table-driven TestUpsertCachedStationBattleOrdering covering the merge edge cases (drops earlier end, replaces
  equal end, drops conflicting future battle, etc.).

  Remaining items

  1. Cache-hit hydration error is still Debug-logged, not returned (station_state.go ~L81). I noted this last round. The author appears to have
  intentionally kept it: TestGetStationRecordReadOnlyRetriesHydrationOnCachedStation exists specifically to assert that a transient hydration
  failure is retried on the next read. So the behavior is deliberate — a transient SELECT failure doesn't poison the station, it just retries.
  I'd bump it to Warn so failures are visible in logs, but this is a nit not a blocker.
  2. Webhook suppression on "all battles just expired." createStationWebhooksWithSnapshot still returns early if currentSignature == "". Net
  effect: when the last active battle expires, no MaxBattle webhook fires for that transition. Likely intentional (the next station-update GMO
  will produce one once is_battle_available toggles), but I'd ask the author to confirm.
  3. Signature-driven save still bypasses the 15-min debounce. Same as last round, but now harmless because the queue coalesces writes, so a
  station whose battles tick across an expiry boundary at most produces one extra coalesced enqueue, not extra DB traffic.
  4. mergeStationBattles is subtle. The "observed is current but not yet active, keep it iff no active battle covers it" branch reads correctly
  against the test cases but is the trickiest function in the file. It's well-tested. Just flagging that future maintenance should be wary of
  edits here.
  5. getKnownStationBattles still does a write-on-read when all battles are expired (collapses to {Loaded:true, Battles:nil}). Bounded and tested
   (TestGetKnownStationBattlesDoesNotMutateCacheOnRead shows it leaves the cache alone when at least one battle is current). Fine.

  Lock semantics, deadlocks, memory

  - All hydration is still under the station mutex, no cross-entity locks introduced.
  - Per-station memory cost unchanged (small slice of battles, typically 1–3 entries).
  - Cache eviction callback still wired correctly via clearStationBattleState.

  Verdict

  A small ask that the author confirm the "all battles expired" webhook suppression is intentional and consider bumping
   the cache-hit hydration error to Warn. The architectural deviations from CLAUDE.md that I flagged previously are gone or are now defensible:
  the entity uses the write-behind queue, batching is real, hydration is lock-protected, and the projection/canonical machinery has collapsed
  into "first element of a sorted slice." The code is meaningfully simpler than the prior revision while doing more.

Mygod:
> I reworked the station_battle path again and the current head should address the earlier architectural concerns:
>
> - station_battle now flushes as a real write-behind batch again, not just coalescing by station_id
> - battle cache state is now a single {Battles, Loaded} map instead of separate cache + hydrated maps
> - the old canonical/projection split is gone; the top battle is just snapshot.Battles[0]
> - station battle projection is now owned only by syncStationBattlesFromProto
>
> On the two remaining behavioral points:
>
> - The currentSignature == "" early return in createStationWebhooksWithSnapshot is intentional for now. That means we do not emit a max_battle webhook on the “last known battle expired” transition.
> - Cache-hit hydration failure in GetStationRecordReadOnly is also intentional for now: on a transient station_battle read failure we still serve the cached station and retry hydration on the next read rather than poisoning the station state. I can bump that log from Debug to Warn later if we want more visibility, but I left the behavior unchanged on purpose.

Fabio1988 (approved):

decoder/api_station.go
```diff
@@ -1,55 +1,55 @@
...
+		Lon:                   station.Lon,
+		Name:                  station.Name,
+		StartTime:             station.StartTime,
+		EndTime:               station.EndTime,
+		IsBattleAvailable:     station.IsBattleAvailable,
+		Updated:               station.Updated,
+		TotalStationedPokemon: station.TotalStationedPokemon,
+		TotalStationedGmax:    station.TotalStationedGmax,
+		StationedPokemon:      station.StationedPokemon,
+		Battles:               battles,
```
jfberry: This should be returning an Api result structure, not the internal structure

decoder/station_battle.go
```diff
@@ -0,0 +1,538 @@
...
+	"github.com/jellydator/ttlcache/v3"
+	"github.com/puzpuzpuz/xsync/v3"
+	log "github.com/sirupsen/logrus"
+
+	"golbat/db"
+	"golbat/pogo"
+)
+
+type StationBattleData struct {
+	BreadBattleSeed           int64      `db:"bread_battle_seed" json:"bread_battle_seed,omitempty"`
```
jfberry: The db data structure is explicitly not the json serialisation structure, Api results and webhooks should have their own field definitions defined specifically for those purposes

decoder/station_battle.go
```diff
@@ -0,0 +1,538 @@
...
+		log.Errorf("Preload: failed to query station battles - %s", err)
+		return 0
+	}
+	defer rows.Close()
+
+	count := int32(0)
+	currentStationId := ""
+	currentBattles := make([]StationBattleData, 0)
+	flushCurrent := func() {
+		storeStationBattles(currentStationId, currentBattles)
```
jfberry: Doesn't this load battles in a way that won't get evicted as their related station entries may not have been loaded into the cache or exist any longer?
Should it verify that the parent station battle is present in memory? Or change the select to join on station

decoder/station_battle.go
```diff
@@ -0,0 +1,538 @@
...
+		if cached.BreadBattleSeed == observed.BreadBattleSeed || cached.BattleEnd <= now || cached.BattleEnd <= observed.BattleEnd {
+			continue
+		}
+		next = append(next, cached)
+	}
+	sortStationBattlesByEnd(next)
+	return next
+}
+
+func getKnownStationBattles(stationId string, now int64) []StationBattleData {
```
jfberry, line 335: I believe that this has to be called under the station lock? If so, should be documented in a comment

decoder/station_state.go
```diff
@@ -147,9 +168,12 @@ func getOrCreateStationRecord(ctx context.Context, db db.DbDetails, stationId st
 
 func saveStationRecord(ctx context.Context, db db.DbDetails, station *Station) {
 	now := time.Now().Unix()
+	battles := getKnownStationBattles(station.Id, now)
+	applyTopStationBattleToStation(station, battles)
```
jfberry, line 190: This fires on every station seen in every GMO, possibly mutating the station object have you run with dbDebugEnabled to ensure that this is not causing accidental dirty station updates and therefore unnescessary writes?
Could applyTopStationBattle return a dirty flag so that the whole list doesn't need to be walked? Perhaps this would mean that oldValues doesn't need to contain the entire Battles array?

decoder/station_state.go
```diff
@@ -252,42 +285,35 @@ func stationWriteDB(db db.DbDetails, station *Station, isNewRecord bool) error {
...
+			Latitude:              station.Lat,
+			Longitude:             station.Lon,
+			Name:                  station.Name,
+			StartTime:             station.StartTime,
+			EndTime:               station.EndTime,
+			IsBattleAvailable:     station.IsBattleAvailable,
+			TotalStationedPokemon: station.TotalStationedPokemon,
+			TotalStationedGmax:    station.TotalStationedGmax,
+			Battles:               battles,
+			Updated:               station.Updated,
```
jfberry: This should be a Webhook version of the battles object

decoder/station_battle.go
```diff
@@ -0,0 +1,538 @@
...
+		if _, ok := seenStations[snapshot.StationId]; !ok {
+			seenStations[snapshot.StationId] = struct{}{}
+			stationIds = append(stationIds, snapshot.StationId)
+		}
+		battles = append(battles, snapshot.Battles...)
+	}
+	return battles, stationIds
+}
+
+func buildDeleteObsoleteStationBattlesQuery(stationIds []string, battles []StationBattleData) (string, []any) {
```
jfberry: Think this is functionally equivelant but matches use of sqlx IN statements elsewhere
``` 
 func buildDeleteObsoleteStationBattlesQuery(stationIds []string, battles []StationBattleData) (string, []any, error) {
      if len(stationIds) == 0 {
          return "", nil, nil
      }
      if len(battles) == 0 {
          return sqlx.In("DELETE FROM station_battle WHERE station_id IN (?)", stationIds)
      }
      seeds := make([]int64, len(battles))
      for i, b := range battles {
          seeds[i] = b.BreadBattleSeed
      }
      return sqlx.In(
          "DELETE FROM station_battle WHERE station_id IN (?) AND bread_battle_seed NOT IN (?)",
          stationIds, seeds,
      )
  }
```

decoder/station.go
```diff
@@ -93,14 +90,11 @@ func (station *Station) Unlock() {
...
 	station.oldValues = StationOldValues{
-		EndTime:                station.EndTime,
-		BattleEnd:              station.BattleEnd,
-		BattlePokemonId:        station.BattlePokemonId,
-		BattlePokemonForm:      station.BattlePokemonForm,
-		BattlePokemonCostume:   station.BattlePokemonCostume,
-		BattlePokemonGender:    station.BattlePokemonGender,
-		BattlePokemonBreadMode: station.BattlePokemonBreadMode,
+		EndTime: station.EndTime,
+		Battles: slices.Clone(battles),
```
jfberry: Combined with my comment on save, this feels like the round trip for read / save (which will happen a lot because of gmos) is not particularly considered.  If I follow the code correctly you don't need a clone of all the battles here, just knowing the top battle seed, and a simple hash of the battles might be more memory efficient and allow for a simpler comparison during save? [and you might not even need the hash if the apply function knew if it had changed anything].
These 'old' values are used to work out if we need a webhook published not for the dirty pathway --> the object itself knows whether it is dirty eliminating the fact that we used to clone objects on read.  The clone object/compare data access model has been re-introduced by this

jfberry (commented):
Needs to be rebased, exclude webhook feature and updated dependencies

jfberry:
Apologies for the delay in review, I needed to have enough time to step through this as it's obviously a fairly significant change from a data processing and call volume perspective.

Mygod:
Rebased on latest main. All comments addressed via the last two commits. Local test in progress.

jfberry:
> Rebased on latest main. All comments addressed via the last two commits. Local test in progress.

thanks, I'll be much faster next time round now I'm familiar with the code.
Only immediate observation so far (courtesy of claude) is hash/fnv is used; hash/maphash is allocation free and faster generally - probably a good match for this use case.  Having learnt this I'm going to swap the other fnv use case (which is much more on the hot path) with this.

Mygod:
> thanks, I'll be much faster next time round now I'm familiar with the code. Only immediate observation so far (courtesy of claude) is hash/fnv is used; hash/maphash is allocation free and faster generally - probably a good match for this use case. Having learnt this I'm going to swap the other fnv use case (which is much more on the hot path) with this.

Fixed. Rust is even allocation freer and faster. ;)

jfberry:
> Fixed. Rust is even allocation freer and faster. ;)

https://github.com/jfberry/golbat-rs

But it's even less understandable. Project abandonned ;-)

jfberry:
I've added a linter as part of the build process, so please rebase on main and fix any issues (I did run a quick scan and actually don't think there are any). I'll then be happy to merge this.

Mygod:
Done.

PR #357, opened 2026-04-28
Bump ringsaturn/tzf to 1.1.0

ringsaturn:
Thanks for using tzf and I have a good news: a new full data with only 17MB can be used since this version.


jfberry:
Thanks very much for your diligence in supporting your end users! (I am assuming you are not also a pokemon scanner).
We'll test this out and adopt it asap

jfberry:
Covered in #361

PR #358, opened 2026-05-01
Add exclude_areas option to webhook config

unseenmagik:
Behavior: 
- Allows a webhook to suppress messages from specific geofences/areas. 
- When an area appears in both areas and exclude_areas, exclusion wins.
- With no areas and only exclude_areas, the webhook receives everything except the excluded areas; with both, exclusion takes precedence over inclusion.


jfberry (approved):

unseenmagik:
Good to merge

PR #359, opened 2026-05-05
Len fort tracker fixes

lenisko:
# Fort tracker: POI removal correctness, hardening, cleanup

Branch: `len-fort-tracker-fixes` — 4 commits.

## Symptom

POIs that should be marked deleted were not removed until Golbat was restarted, and often not even then. A restart only "fixed" it if the first GMO post-restart contained full coverage
of the cell — partial first GMOs left the wrong fort set in tracking.

## Root cause

Two interacting bugs in `ProcessCellUpdate`'s `firstScan` branch and in preload bookkeeping:

1. `cell.lastSeen` was only ever written inside `ProcessCellUpdate`. Preload (`Preload` / `LoadFortsFromDB` / `RegisterFort`) never set it. After every restart, every cell had
`cell.lastSeen == 0`, so the next GMO took the `firstScan` path.
2. The `firstScan` branch overwrote `cell.pokestops` / `cell.gyms` with the contents of that single GMO. Any preloaded fort not in that one GMO was dropped from per-cell tracking.
Because stale detection iterates `cell.pokestops` / `cell.gyms`, those forts could never be flagged stale again — they were stuck as "live" entries in `ft.forts` pointing at a cell that
 no longer listed them.

A restart only helped when the next first-GMO happened to enumerate every fort in the cell.

## Other bugs found and fixed

- **`RestoreFort` units bug.** Callers passed `time.Now().Unix()` (seconds) into a milliseconds field. `lastSeen` came out ~1000× too small; the very next GMO saw a ~1e12 ms gap, blew
past the 1h stale threshold, and re-staled the fort immediately.
- **`RestoreFort` wall-clock vs GMO timestamp drift.** After fixing the units bug, callers in `updatePokestopFromFort` / `updateGymFromFort` still used `time.Now().UnixMilli()` instead
of the GMO's own `fort.Timestamp`. On replay or ingest lag the restored fort's `lastSeen` ran ahead of (or behind) the cell's `lastSeen`, which is sourced from the GMO timestamp —
leaving staleness math inconsistent across the cell and its restored member.
- **Future-dated `AsOfTimeMs` wedge.** A clock-skewed or malicious scanner could permanently silence a cell — every later real GMO had `timestamp <= cell.lastSeen` and was dropped at
the early-return.
- **`clear*WithLock` ghost entries.** When the gym/pokestop could not be loaded from cache or DB, the tracker entry was left in place. Same fort was re-selected as stale on every
subsequent GMO.
- **Dangling `UpdateFort`.** Only caller was `RestoreFort`, and the units bug rode through it.
- **GMO type-change race.** If the same id appeared in `pokestopIds` and `gymIds` of one GMO, it was inserted into both sets and the conversion logic flipped back and forth.
- **Preload seeded `lastSeen` from `pokestop.Updated`.** Once commit 1 made `cell.lastSeen` non-zero post-preload (so stale detection actually ran), this became live: any fort whose DB
`Updated` predated the stale threshold was flagged stale on the first GMO that didn't include it. Co-primary with the root cause; fixed together in commit 3.

## Fix summary

Four commits, in order:

### 1. `fix(fort_tracker)` — `d20d5ca`

- `RegisterFort` seeds `cell.lastSeen`, so the first post-restart GMO is no longer treated as a `firstScan`.
- `firstScan` branch merges (does not replace) cell fort sets, refreshing `lastSeen` for retained forts.
- Reject GMOs with timestamps more than 60s in the future.
- Dedup ids that appear in both `pokestopIds` and `gymIds` (gym wins, matches proto pack order).
- Defensive insertion into the new cell set during cell-move handling.
- `clear*WithLock` removes the tracker entry when the entity cannot be loaded.
- Drop unused `UpdateFort`. `RestoreFort` does the work directly with explicit ms units; callers now pass `time.Now().UnixMilli()`.

### 2. `refactor(fort_tracker)` — `11f8a04`

- Collapse `loadPokestopsFromDB` / `loadGymsFromDB` into one `loadFortKindFromDB` helper.
- Replace `clearGymWithLock` / `clearPokestopWithLock` with a generic `clearFortWithLock` plus per-kind ops tables (`gymClearOps`, `pokestopClearOps`).
- Hoist debug "pending removal" logs out of the global tracker lock by splitting `ProcessCellUpdate` into a thin wrapper plus locked core that returns the pending lists.
- Downgrade per-conversion log to Debug (already logged at Info by `clearFortWithLock`).
- Iterate `cellForts` directly in `CheckRemovedForts`; drop the parallel `cellsToBeCleaned` slice in `decodeGMO`.
- `strconv.FormatUint` for cellId formatting in API responses (S2 cell ids exceed JS 2^53 safe-int — observed DB max 8935267727656878080 — so they're stringified for JSON consumers).
- Fix doc-comment / type-name mismatches on `FortTrackerLastSeen` and `FortTrackerGMOContents`.

### 3. `feat(fort_tracker)` — `26140b2`

- New config `cleanup.forts_min_miss_count` (default `1`, preserves prior behaviour). When set higher, a fort must be absent from N consecutive cell scans past the time threshold before
 being flagged stale — defends against transient single-frame coverage gaps and level-30 gating.
- Track `missCount` per fort, reset on every sighting (including the `firstScan` merge path).
- Treat preload as a confirmation event: `lastSeen` is `time.Now().UnixMilli()` instead of `row.Updated * 1000`. Required for commit 1's `cell.lastSeen` seeding to be safe — otherwise
preloaded forts with old `Updated` values get instantly staled.
- Tests for the §6.6 design-review scenarios:
  - partial first GMO post-preload preserves preloaded forts;
  - future-dated GMO does not wedge the cell;
  - `RestoreFort` followed by partial GMO does not re-stale;
  - id present in both id-lists dedups (gym wins);
  - `missCount` guard with N=2 needs two consecutive misses;
  - `missCount` resets on intervening sighting.

### 4. `fix(fort_tracker)` — RestoreFort uses GMO timestamp

- `updatePokestopFromFort` passes `now*1000` (param already received as `fort.Timestamp/1000` seconds) into `RestoreFort` instead of `time.Now().UnixMilli()`.
- `updateGymFromFort` gains a `timestampMs int64` parameter; `gmo_decode.go` call site passes `fort.Timestamp`.
- Eliminates drift between a restored fort's `lastSeen` and its cell's `lastSeen` under ingest lag or replay.

## Out of scope

- **Unbounded growth of `ft.cells` / `ft.forts`.** GMOs sometimes return empty; we don't want to drop cells whose forts vanished transiently.
- **Race between `LoadFortsFromDB` and live ingest.** Loading already blocks before listeners accept traffic.
- **Soft-delete intermediate state in DB (§6.3).** We don't want to change the DB-side stale-to-deleted transition here.

## Testing

- `go test ./...` — all green.
- Tracker tests cover all five §6.6 scenarios from the design review, plus regression tests for `missCount` semantics.

## Config

```toml
[cleanup]
forts_stale_threshold = 3600   # seconds, default 1 h (unchanged)
forts_min_miss_count  = 1      # consecutive cell-scan misses, default 1

# Operators wanting fewer false positives from transient coverage gaps can set forts_min_miss_count = 2 or higher.
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -57,7 +57,8 @@ type cleanup struct {
 	Stats               bool  `koanf:"stats"`
 	StatsDays           int   `koanf:"stats_days"`
 	DeviceHours         int   `koanf:"device_hours"`
-	FortsStaleThreshold int64 `koanf:"forts_stale_threshold"` // seconds, default 3600 (1 hour)
+	FortsStaleThreshold int64 `koanf:"forts_stale_threshold"`  // seconds, default 3600 (1 hour)
+	FortsMinMissCount   int   `koanf:"forts_min_miss_count"`   // consecutive cell-scan misses before staleness (default 1)
 }
 
 type Webhook struct {
--- a/decode.go
+++ b/decode.go
@@ -483,24 +483,20 @@ func decodeGMO(ctx context.Context, protoData *ProtoData, scanParameters decoder
 	var newNearbyPokemon []decoder.RawNearbyPokemonData
 	var newMapPokemon []decoder.RawMapPokemonData
 	var newMapCells []uint64
-	var cellsToBeCleaned []uint64
 
-	// track forts per cell for memory-based cleanup (only if tracker enabled)
+	// track forts per cell for memory-based cleanup (every map cell gets an
+	// entry, so empty fort lists are seen as "no forts" by the tracker)
 	cellForts := make(map[uint64]*decoder.FortTrackerGMOContents)
 
 	if len(decodedGmo.MapCell) == 0 {
 		return "Skipping GetMapObjectsOutProto: No map cells found"
 	}
 	for _, mapCell := range decodedGmo.MapCell {
-		// initialize cell forts tracking for every map cell (so empty fort lists are seen as "no forts")
 		cellForts[mapCell.S2CellId] = &decoder.FortTrackerGMOContents{
 			Pokestops: make([]string, 0),
 			Gyms:      make([]string, 0),
 			Timestamp: mapCell.AsOfTimeMs,
 		}
-		// always mark this mapCell to be checked for removed forts. Previously only cells with forts were
-		// added which meant an empty fort list (all forts removed) was never passed to the tracker.
-		cellsToBeCleaned = append(cellsToBeCleaned, mapCell.S2CellId)
 
 		if isCellNotEmpty(mapCell) {
 			newMapCells = append(newMapCells, mapCell.S2CellId)
@@ -562,7 +558,7 @@ func decodeGMO(ctx context.Context, protoData *ProtoData, scanParameters decoder
 	}
 
 	if scanParameters.ProcessGyms || scanParameters.ProcessPokestops {
-		decoder.CheckRemovedForts(ctx, dbDetails, cellsToBeCleaned, cellForts)
+		decoder.CheckRemovedForts(ctx, dbDetails, cellForts)
 	}
 
 	newFortsLen := len(newForts)
--- a/decoder/fort_tracker.go
+++ b/decoder/fort_tracker.go
@@ -24,6 +24,7 @@ type FortTracker struct {
 
 	// Configuration
 	staleThreshold int64 // seconds after which a fort is considered stale
+	minMissCount   int   // consecutive cell scans a fort must be missing before staleness
 }
 
 // FortTrackerCellState tracks the state of forts within a single S2 cell
@@ -35,9 +36,10 @@ type FortTrackerCellState struct {
 
 // FortTrackerLastSeen holds tracker bookkeeping (cell + last-seen + type) for a fort
 type FortTrackerLastSeen struct {
-	cellId   uint64
-	lastSeen int64 // last time this fort was seen in GMO
-	isGym    bool  // faster targeted removal
+	cellId    uint64
+	lastSeen  int64 // last time this fort was seen in GMO
+	missCount int   // consecutive cell scans where this fort was absent
+	isGym     bool  // faster targeted removal
 }
 
 // FortTrackerGMOContents holds fort IDs per cell extracted from a single GMO
@@ -50,14 +52,20 @@ type FortTrackerGMOContents struct {
 // Global fort tracker instance
 var fortTracker *FortTracker
 
-// InitFortTracker initializes the global fort tracker
-func InitFortTracker(staleThresholdSeconds int64) {
+// InitFortTracker initializes the global fort tracker.
+// minMissCount must be >= 1 (1 = previous behavior, >1 = require multiple
+// consecutive cell scans missing the fort before flagging it stale).
+func InitFortTracker(staleThresholdSeconds int64, minMissCount int) {
+	if minMissCount < 1 {
+		minMissCount = 1
+	}
 	fortTracker = &FortTracker{
 		cells:          make(map[uint64]*FortTrackerCellState),
 		forts:          make(map[string]*FortTrackerLastSeen),
 		staleThreshold: staleThresholdSeconds,
+		minMissCount:   minMissCount,
 	}
-	log.Infof("FortTracker: initialized with stale threshold of %d seconds", staleThresholdSeconds)
+	log.Infof("FortTracker: initialized with stale threshold of %d seconds, min miss count %d", staleThresholdSeconds, minMissCount)
 }
 
 // LoadFortsFromDB populates the tracker from database on startup
@@ -114,6 +122,9 @@ func loadFortKindFromDB(ctx context.Context, dbDetails db.DbDetails, table strin
 			break
 		}
 
+		// "now" rather than row.Updated: load is a confirmation event.
+		// See preloadPokestops for rationale.
+		nowMs := time.Now().UnixMilli()
 		fortTracker.mu.Lock()
 		for _, row := range rows {
 			cellId := uint64(row.CellId)
@@ -123,9 +134,12 @@ func loadFortKindFromDB(ctx context.Context, dbDetails db.DbDetails, table strin
 			} else {
 				cell.pokestops[row.Id] = struct{}{}
 			}
+			if nowMs > cell.lastSeen {
+				cell.lastSeen = nowMs
+			}
 			fortTracker.forts[row.Id] = &FortTrackerLastSeen{
 				cellId:   cellId,
-				lastSeen: row.Updated * 1000, // convert to milliseconds
+				lastSeen: nowMs,
 				isGym:    isGym,
 			}
 		}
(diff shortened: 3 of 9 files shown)


lenisko:
Might need some long-term testing, cc @Fabio1988

jfberry:
Some parts of the description don’t make any sense

lenisko:
I will review and slap my tool later on :)

lenisko:
Updated. Haven't tested myself yet.

lenisko:
Started testing

lenisko:
So far results are far better. @Fabio1988 mind joining in testing?

jfberry:
Reviewed on my phone but I saw: cellid as string in data structure? Why?
Restoring forts from gmo takes now milliseconds, not time from gmo?

lenisko:
> Reviewed on my phone but I saw: cellid as string in data structure? Why? Restoring forts from gmo takes now milliseconds, not time from gmo?

It's uint64 internally, string for JSON debug output. I pulled max value of `8935267727656878080` on my end, which is past what JS `int` can get.

Fixed passed ts.

Fabio1988 (approved):
tested, works well so far

lenisko:
Been working on it a while, no issues observed anymore. Tested with restart and no forts been removed/converted straight after reboot (which was the case before).

jfberry:
Read through under `-race`. Three issues worth flagging.

## 1. Type-conversion orphan in `applyPresentForts` (pre-existing, but this PR widens reachability)

When `info.isGym != isGym`, we flip `info.isGym` and add to the new cell set but never remove from the OLD cell set ([`fort_tracker.go:393-403`](https://github.com/UnownHash/Golbat/blob/295214e712051e8f743ac8986d1c45227fb2a6b9/decoder/fort_tracker.go#L393-L403)):

```go
if info.isGym != isGym {
    info.isGym = isGym
    *converted = append(*converted, id)
    ...
}
```

Repro trace:
1. `RegisterFort("X", cell, isGym=false, t0)` — `cell.pokestops[X]` set, `info.isGym=false`.
2. GMO at `t1` with X **only** in `gymIds`. After dedup, `currentGyms={X}`, `currentPokestops={}`.
3. `applyPresentForts` for gyms flips `info.isGym=true`, sets `cell.gyms[X]`, **does not** touch `cell.pokestops[X]`.
4. Staleness loop on `cell.pokestops` sees X (still there), not in `currentPokestops`, `info.missCount=1`, `missingDuration=0` → pending. Pending re-adds X to `currentPokestops`. `cell.pokestops = currentPokestops` keeps X.

End state: X is in both `cell.pokestops` and `cell.gyms`, with `info.isGym=true`.

Eventually the orphan trips a stale-pokestop on a row that's already marked deleted. `RemoveFort` is then called, and its choice of which cell set to prune is driven by `info.isGym` — now `true` — so it deletes from `cell.gyms` (the live entry) and leaves the orphan in `cell.pokestops`. Next GMO re-creates the gym side; cycle resumes.

This code path exists on `main` too (confirmed by reading `main:decoder/fort_tracker.go`), so it's pre-existing. But the firstScan fix in this PR makes it more reachable in real ops, because more forts now survive the partial-first-GMO window long enough for a type conversion to land on them.

Fix mirrors the existing cell-move cleanup pattern at lines 381-389:

```go
if info.isGym != isGym {
    if info.isGym {
        delete(cell.gyms, id)
    } else {
        delete(cell.pokestops, id)
    }
    info.isGym = isGym
    *converted = append(*converted, id)
    ...
}
```

Worth a focused unit test: "fort observed as gym after being registered as pokestop leaves `cell.pokestops` empty."

## 2. `RestoreFort` doesn't clean isGym-mismatched old entry (defensive)

[`fort_tracker.go:445-453`](https://github.com/UnownHash/Golbat/blob/295214e712051e8f743ac8986d1c45227fb2a6b9/decoder/fort_tracker.go#L445-L453):

```go
if existing, exists := ft.forts[fortId]; exists && existing.cellId != cellId {
    if oldCell, ok := ft.cells[existing.cellId]; ok {
        if existing.isGym { delete(oldCell.gyms, fortId) } else { delete(oldCell.pokestops, fortId) }
    }
}
```

Only cleans the old cell when the cell changed. If only `isGym` changed in the same cell — e.g. a fort that was a gym, got marked deleted (still in the tracker because conversion uses `removeFromTracker=false`), then later comes back as a pokestop via `updatePokestopFromFort` — the old-type set isn't cleared. Today's callers don't reach this because `pokestop_decode.go` always passes `isGym=false` and `gym_decode.go` always passes `isGym=true`, but the assumption is fragile.

Defensive fix:

```go
if existing, exists := ft.forts[fortId]; exists && (existing.cellId != cellId || existing.isGym != isGym) {
    ...
}
```

## 3. Stale-return → `clearFortWithLock` race (pre-existing)

`ProcessCellUpdate` releases `ft.mu` between deciding X is stale and `clearFortWithLock` actually marking it deleted. If a new GMO arrives in that window and re-observes X, `applyPresentForts` re-adds X and updates `info.lastSeen`; then `clearFortWithLock` marks the DB row deleted and `RemoveFort` prunes the just-restored entry. The next GMO triggers un-delete + `RestoreFort` and the system stabilises. End state correct, but consumers see a REMOVAL → NEW webhook flap. Same behaviour exists on `main`; not a regression.

## Lock ordering

Two paths cross `ft.mu` and entity locks:
1. `updatePokestopFromFort`/`updateGymFromFort` → `RestoreFort` — holds entity lock, then `ft.mu`.
2. `CheckRemovedForts` → `clearFortWithLock` — `ft.mu` acquired in `ProcessCellUpdate`, **released**, then entity lock acquired, then `ft.mu` re-acquired in `RemoveFort`.

Consistent ordering: entity → `ft.mu`. The unlock in `ProcessCellUpdate` before `clearFortWithLock` acquires the entity lock is load-bearing — a future refactor that moves the unlock inside would deadlock. Worth a one-line comment on `processCellUpdateLocked`.

## Minor

- `processCellUpdateLocked` is order-sensitive: `firstScan := cell.lastSeen == 0` MUST be captured before `cell.lastSeen = timestamp`. A one-line comment would help.
- `pendingPokestops` / `pendingGyms` are returned out of the lock for log formatting; this is safe because string content is immutable. Worth a note explaining why.

lenisko:
@jfberry done https://github.com/UnownHash/Golbat/pull/359/changes/a6027d1cf9feb974e538bdd4f702d3ae8fa9c4bc

lenisko:
Fresh changes tested on prod since yesterday, with no issues observed. 

Would like to get it asap to main since old version had holes and couldn't handle all cases during run-time but only after Golbat restart.

jfberry:
Yep, just confirm that it's working still and then we can do a squash merge onto main

lenisko:
Confirmed

PR #360, opened 2026-05-19
feat: add super mega eligible field to gym

Fabio1988:
only some gyms are eligible for super mega raid


jfberry:
Looks good to me, assuming tested

jfberry:
Has this now started being populated?

Fabio1988:
Checked my DB, wondering why not, as falink was live yesterday .. also checked during live event yesterday 

```
select distinct(is_mega_enhanced_eligible) from golbat.gym;
0
```

Fabio1988:
with date of today, still not populated. I will check on the weekend again

PR #363, opened 2026-05-25
Read raid info from GymInfo proto

jfberry:


Mygod:
Also defender update should probably also update team_id and available_slots?

jfberry:
> Also defender update should probably also update team_id and available_slots?

It does from the fort info embedded update

PR #376, opened 2026-07-02
feat: add raid seed, cpm and stamina to gym records + alignment to raid_stats

ccev:
- raid seed: acts as a id for raids, also useful in other raid-related requests
- cpm and stamina are useful for technical players. i got asked by people to show those in diadem
- alignment for raid_stats to support shadow raids


jfberry:
This will need updating and retesting - stats have been split off into a queue with non-locked message passing so it may require additional values to be sent over the queue (or not, I can't remember if this was only pokemon or all stats).

jfberry:
I wonder whether alignment is needed for raid stats given that level gives you the same information?

ccev:
Indeed, but that would be guessing for data that we can get from the game. You're expecting consumers to update their apps with every new alignment or raid level that may have special alignments.

PR #377, opened 2026-07-04
Pipeline smoothing: no global lock points on the hot path, graceful degradation under overload

jfberry:
## Thesis

**Smooth the pipeline: no global lock points on the hot path, and graceful degradation when any subsystem overloads.** Every change on this branch serves one of those two clauses. Before: any stall — a cache expiry sweep, a DB blip, a geofence reload, a saturated worker — could freeze all ~96 decode goroutines behind a shared lock, and overload responses were unbounded (goroutine pileups measured in the millions). After: hot-path work is lock-free or single-owner, every cross-subsystem handoff is a bounded queue, and each queue's overflow policy matches its data's loss tolerance — loss-tolerant paths drop and count, loss-intolerant paths apply bounded backpressure. Stalls now cost latency in one subsystem instead of freezing the pipeline.

Developed and verified against the heaviest known production instance (~15 cores, 10M-entry caches, 16k webhooks/s) across a week of live incidents. Every mechanism below was named by a goroutine dump, CPU profile, or production log before its fix shipped; several plausible theories were falsified by the same evidence and are documented in commit messages as dead ends.

## The hot path, before → after

| Global lock point (before) | Mechanism (after) |
|---|---|
| Pokemon R-tree mutex: savers, per-eviction goroutines, and per-request tree copies convoyed (dumps: 90–96/96 decoders parked; 5–25s entity locks) | **Single ordered tree writer** (`decoder/rtree_evictor.go`): all runtime mutations flow through one batching worker; scans use a 1s shared snapshot; savers never touch the mutex |
| `pokemonStatsLock` under entity locks (66/96 parked) | **Stats aggregation worker**: snapshot events over a channel; one lock acquisition per 512-event batch |
| FortTracker mutex (27/96 parked) | **Fort tracker worker**: staleness processing off the decode path, per-event DB deadlines |
| Cache expiry sweeps + per-eviction goroutines (ttlcache): synchronized cohorts froze savers; eviction callbacks spawned unbounded goroutines — one incident parked **3.17M goroutines holding entity locks** | **otter v2 across all caches** behind a hardened adapter (`cache/`): lock-free reads, timing-wheel expiry, eviction events on one bounded dispatcher per cache. ttlcache removed from go.mod |
| Geofence matching inside entity-lock critical sections (33.7% of box CPU; orb re-derived ring bounds per call) | **Compiled geofences** (cached bounds, 5×) + **S2 cell lookup** fast path (coarse interior covering, hole-aware, async hash-gated rebuilds) |
| Spawnpoint entity lock once per wild pokemon (readers queued 0.4–1.2s behind writers holding it across DB loads) | **Atomic mirrors** (despawn second, last-seen): the per-pokemon despawn check and the per-sighting no-change update are lock-free end to end |
| Scan lookup cache: two pointers per entry = 3 dependent DRAM misses × 15–20M loads/s | **Inline-by-value lookup entries** + hoisted scan-loop copy (zero allocations per candidate) |
| Unbounded goroutine per raw request | **Bounded ingest** (`raw_processing_concurrency`, default min(4×CPU, 96)) with a deep parked queue and aggregated shed accounting |

## Graceful degradation: the never-block invariant

No decode-path producer may block on a worker whose throughput can fall below the event rate (two production incidents — a 30s fill-drain limit cycle and the 3.17M-goroutine bomb — were both blocking sends). Documented in CLAUDE.md "Decode-Path Workers":

| Handoff | Full-queue behavior | Rationale |
|---|---|---|
| Tree writer (savers) | block (bounded producers: ≤96) | a lost insert = invisible live pokemon |
| Tree writer (evictions) | drop + count | a lost delete = ghost point; scans verify against the lookup cache |
| Stats events | drop + count | statistics are loss-tolerant; decode is not block-tolerant |
| Fort tracker | drop + count | staleness needs an hour of absence; one observation is worthless |
| Cache eviction dispatch | drop + count (the one non-self-healing loss — has its own counter) | blocking would stall otter maintenance |
| Raw ingest | shed + count past 32× queue | bounded loss under overload beats unbounded pileup |

An independent audit of all six judged every worker load-bearing (three map 1:1 to goroutine-dump-verified convoys; the dispatcher is structurally forced by otter's execution model) — findings and parameter fixes in commit 8cb9241.

## Production verification

Baseline vs final, 60s CPU windows at comparable load (full three-way table in `docs`/PR history):

- Entity-lock freezes: **5–25s → worst wait 1.48s** (external DB ripple, absorbed without loss); hour-boundary windows that previously shed 15k packets/s now pass with zero limiter events and ~1µs scan lock times
- ttlcache machinery (~7–8% CPU): **eliminated**; otter costs ~4.5–5% doing the same job; the #177 write-contention risk did not materialize (0.93%)
- Heap: **43.4 → 36.3 GB**; goroutines **166 → 93**; GC mark share 20.7% → 16.9%
- Spawnpoint DB SELECTs: **2.1k/s → ~0.6/s** (removed churn — a hysteresis workaround's blind spot, obsoleted by otter's true touch-on-hit)
- Scan treatment band: **−15%** from the lookup layout change (~3.5% total CPU; bench predicted more — transfer gap documented in `cachebench/lookup_bench_test.go`)
- Shutdown preservation: parallelized (~13k → ~50k+ rows/s) with total/rate/ETA logging; **operators must set the process manager kill window above the preserve duration** (pm2 `kill_timeout`)

## Observability added

Per-caller DB timing (`[DB_SLOW]` + `golbat_db_query_duration_seconds{caller}` histogram with pool-starvation attribution), worker backlogs (`golbat_worker_backlog{worker}`), drop counters for every loss path that cannot self-heal, aggregated once-per-second drop/shed logging (log storms amplified two incidents), `TrackedMutex` holder/waiter naming. The instrumentation repeatedly out-diagnosed CPU profiles — lock waits are invisible to on-CPU sampling.

## Also fixed en route

Cache construction ran at package init before config load (`fort_in_memory`'s TTLs and fort eviction callbacks had never been active); Koji outage at boot crash-looped startup; geofence reload races (stale-forever content gate, mixed-epoch S2 publication); S2 hole rings; fort tracker DB calls without deadlines; ttlcache bumped v3.4.1 before removal. Two adversarial multi-agent review waves ran over the branch; all findings fixed in-branch, several production-grade.

## Config / operator notes

- New tuning: `raw_processing_concurrency`, `raw_processing_queue_factor`, `slow_db_query_ms`; `write_behind_worker_count` default 50 → 16 (burst-amplification evidence in commit f36a73e); `cache_shards` is obsolete and ignored
- `pm2 kill_timeout` ≥ preserve duration for `preserve_pokemon` instances
- Recommended DB-side (not in this PR): `innodb_log_buffer_size ≥ 512M`; identify the quarter-hourly job behind residual `DB_SLOW` ripples
- Benchmark harness + cache investigation committed under `cachebench/` + `docs/cache-investigation-brief.md` as the regression gate and design record


jfberry:
## Post-review fix wave (max-effort review findings)

A deep multi-agent review of this branch surfaced 15 verified findings; the last three commits address them. The most important correction first:

**The original design premise was wrong.** ttlcache v3 runs `OnEviction` callbacks on their **own goroutines** (`Cache.OnEviction` wraps callbacks in `go fn(...)`) — they were never synchronous with the expiry sweep or shard lock. The production freezes were the **global tree-mutex convoy**: mass expiry spawned thousands of callback goroutines all contending for `pokemonTreeMutex`, amplified by permanent copy-on-write from per-request tree copies, with savers (holding entity locks) queued behind them. The batching/snapshot work still targets exactly that convoy — but the async callbacks needed race guards this branch initially lacked:

- **Pokemon eviction vs. concurrent update race** (live pokemon vanishing from all scans): the callback now takes the entity lock and skips cleanup if the pokemon was re-cached; saves self-heal by re-inserting the tree point when an eviction removed the lookup mid-update.
- **Fort eviction guards**: skip when the lookup entry is already gone (deleted fort — prevents unpaired deferred deletes erasing restored forts) or owned by a different fort type (pokestop↔gym conversion — prevents the stale counterpart's expiry from wiping the live fort).
- **Fort TTL jitter** (25–27h / 60–70min per entry): note that fixing the init-ordering bug *activates* the 25h fort TTL + eviction callbacks for the first time; without jitter the entire preload cohort would mass-expire at T+25h.
- **Restart ghost pokemon**: `PreloadPreservedPokemon` now skips unverified pokemon explicitly (the old `DefaultTTL=0` sentinel used to do this implicitly; jittered TTLs broke that).
- **Stats dedup window**: the encounter-dedup cache keeps its full 60-min window instead of inheriting the pokemon cache's 1-minute past-despawn clamp.
- **Raw limiter OOM guard**: parked decode goroutines (payloads pinned, ingest already returned success) are now capped at 8× the concurrency limit; beyond that packets are shed with a warning.
- Remaining tree-lock-under-shard-lock inversion removed from the rehydration `GetOrSetFunc` closures; `ProactiveIVSwitch` uses a fresh tree copy; scan paths dedupe matched ids; CLAUDE.md and the plan doc now describe the correct concurrency model.

Full `go test ./... -race`, vet, build green.

jfberry:
## Benchmark: which pieces are actually load-bearing?

We traced tidwall/rtree v1.10.0's COW internals (64-entry nodes, `cow()` clones each touched node once per `Copy()` generation) and benchmarked the two contested mechanisms against a 500k-point tree with 8 concurrent savers doing lock+insert (the Golbat saver pattern — entity lock held across the tree write). Ratios are the signal; absolute numbers are a 12-core laptop.

**Mass-eviction drain — per-eviction goroutines (main's behavior: ttlcache spawns one per evicted item) vs the batched evictor worker (512 deletes per lock hold):**

| config | drain | saver lock-wait p99 | max |
|---|---|---|---|
| convoy, 200k evictions, scans@50ms | 1.06 s | **948 ms** | 948 ms |
| batched, 200k evictions, scans@50ms | 645 ms | **19.8 ms** | 25 ms |
| convoy, 200k evictions, no scans | 315 ms | 309 ms | 309 ms |
| batched, 200k evictions, no scans | 75 ms | 3.8 ms | 4.1 ms |

**Verdict: the batched evictor is unambiguously load-bearing** — 40–75× reduction in saver tail latency in every configuration. This is the mechanism behind the production `LOCK_HELD_LONG` freezes, and batching kills it.

**COW write tax — how often the tree gets `Copy()`d:**

| copy frequency | write cost | allocs /100k writes |
|---|---|---|
| never | 561 ns | ~1 MB |
| every 1000 writes (≈1s snapshot) | 2.4 µs | 2.27 GB |
| every write (per-request) | 2.5 µs | 2.27 GB |

**Honest correction to this PR's description: the clone tax *saturates*.** Random writes touch fresh leaves every generation, so once copies happen at all regularly, 1s-shared snapshots cost nearly the same per-write as per-request copies. The snapshot change does **not** meaningfully reduce steady-state write cost, contrary to the "amortizes COW" framing. Its real (and still worthwhile) benefits: scans become lock-free — under a writer convoy, per-request `RLock`+`Copy` would queue behind the entire writer backlog (seconds of API latency), and scan RLock interleaving also measurably stretches drains (1.06s vs 918ms); plus zero copies happen when no scan traffic arrives. ~40 lines, low complexity, kept — but for those reasons, not the retracted one.

TTL jitter attacks a different blocker entirely (ttlcache's own shard-locked sweep unlink work, proportional to cohort size) and is unaffected by these measurements.

PR #378, opened 2026-07-06
Phase 0: raw payload capture + protobench decode-at-volume harness

jfberry:
## What this is

Phase 0 of the proto-decoding GC-reduction effort (successor to #323, redone measurement-first). It proves — or kills — the opaque/lazy decode methodology **before** any migration of Golbat's ~31 pogo-touching files happens.

Design spec: `docs/superpowers/specs/2026-07-05-proto-decoding-gc-design.md`
Implementation plan: `docs/superpowers/plans/2026-07-06-phase0-capture-and-protobench.md`

## What's included

**1. Raw payload capture hook (Golbat, config-gated, off by default)**
- `[raw_capture]` config: samples raw proto payloads on the decode path into `capture/<METHOD>/<unixMilli>_<size>.bin`
- Size-stratified quotas (200 per size bucket × 5 buckets = max 1,000/method) so rare complex shapes (raid-heavy GMOs, invasions, stations) make it into the corpus
- Non-blocking producer (drop + rate-limited count when the channel is full — per the decode-path invariant); quota state survives restarts; same-ms filename collisions uniquified

**2. `protobench/` — standalone decode-at-volume harness (separate Go module)**
- Own pogo package generated at `API_HYBRID` with `[lazy = true]` annotations (88 fields, incl. repeated message fields — the v1.36 runtime supports them). Separate module is mandatory: duplicate vbase.proto registration in one binary panics the proto registry. Generated code is gitignored; `scripts/gen.sh` is the one-command regen (protoc-gen-go pinned v1.36.11)
- Three configurations from one codebase: **(a)** default build = open layout (current-Golbat baseline), **(b)** `-tags protoopaque` = opaque + lazy, **(c)** `-tags protoopaque -nolazy` = opaque without lazy
- Access-simulation readers (GMO, Encounter) mirroring the fields Golbat's decoder actually reads, so lazy's pay-on-access cost is measured honestly
- Per-method microbenchmarks + a sustained volume runner (default 96 workers) reporting decodes/s, alloc/decode, GC CPU share, GC pause p50/p99 via `runtime/metrics`, with an optional pointer-dense heap ballast

## Exit gate (from the spec)

Config (b) or (c) must beat (a) at volume on allocation rate and GC CPU share — provisional target ≥20% allocs/op reduction on GET_MAP_OBJECTS. If the gate fails, the Phase 1 migration does not happen and only this instrumentation merges.

## Verification

- Root module: `go build ./... && go test ./...` clean
- protobench: `go vet` + `go test` clean in **both** build modes
- Multi-stage review: per-task spec+quality reviews, whole-branch final review; findings fixed (percentile math edge cases, annotation-scan self-reference, write-failure log flooding, filename collisions)


jfberry:
## Decode-engine experiment results (frozen prod corpus, 800 GMO + 200 ENC)

The Phase 0 opaque/lazy gate **failed** (lazy adds bookkeeping cost on GMO's many tiny unread subtrees), so the harness was pointed at alternative engines. All engines are equivalence-tested — identical access-walk results over every corpus payload.

**GMO, per-decode (12 workers, no ballast):**

| engine | decodes/s | B/decode | objects | GC CPU |
|---|---|---|---|---|
| std (current) | 34.4k | 222KB | 1,357 | 5.1% |
| opaque+lazy | 31.2k | 221KB | 2,043 | 4.8% |
| vtprotobuf+pools | 55.7k | 71KB | 646 | 4.9% |
| **hyperpb (+PGO, arena reuse)** | **58.6k** | **50KB** | **255** | **1.4%** |

**With a 2GB pointer-dense ballast (prod-shaped live heap) the gap widens to 3.2–3.6×** — std collapses to 11.6k/s while hyperpb/vtpool hold ~37–41k/s. This is the regime prod runs in.

The Golbat-reusable piece: `cmd/hyperpbgen` generates typed, protoc-style accessor shims over hyperpb (nil-chaining-safe, float32-exact) — the shim layer costs ~1.7%. Engine-independent cheap wins also measured: `DiscardUnknown` (+3.7% rate), ingest buffer pooling (−14% bytes), GOGC 200–400.

Full write-up and migration recommendation: `protobench/RESULTS.md`.

jfberry:
## How hyperpb would land in Golbat (migration analysis)

Golbat is strictly read-only against client protos, which removes hyperpb's main constraint (no mutation). The migration shape:

**Parse sites (decode.go).** Compile `hyperpb.MessageType`s once at startup; each decode borrows a `hyperpb.Shared` arena, parses, wraps in generated shims, frees the arena when the (synchronous) processing completes:

```go
shared := gmoArenas.Get().(*hyperpb.Shared)
defer func() { shared.Free(); gmoArenas.Put(shared) }()
msg := shared.NewMessage(gmoType.Load())
if err := msg.Unmarshal(sDec); err != nil { ... }
gmo := pogoshim.AsGetMapObjectsOutProto(msg.ProtoReflect())
```

**Field reads switch to generated shim getters** (`fortData.FortId` → `fort.GetFortId()`, `range gmo.MapCell` → `range gmo.GetMapCell().All()`) — mechanical, compile-checked, produced by `protobench/cmd/hyperpbgen` pointed at `golbat/pogo` descriptors so existing enum types keep compiling. Entities already copy values out of protos, so nothing downstream of the decode functions changes. Measured shim overhead: ~1.7%.

**Three boundaries need explicit handling:**
1. `weather_consensus.go` retains `*pogo.ClientWeatherProto` for hours → store a small value struct of the ~8 fields it reads.
2. Raw batch structs hold shims (die before `Free`) — assert nothing stashes them.
3. hyperpb is zero-copy: the payload buffer must outlive the arena — trivially true today; couples with future ingest buffer pooling via a shared release point.

**Insulation:** call sites depend on generated getters we own, not hyperpb's v0.1.x API — only the parse glue touches hyperpb. Per-method, flag-gated rollout with shadow comparison against std (easy because read-only).

**Further headroom identified:** ScanParameters-driven subtree skipping becomes real under lazy parsing (disabled cells ≈ never decoded); prod PGO recompile (+4%); per-worker arena affinity; DiscardUnknown / GOGC 200–400 / ingest pooling as engine-independent wins; re-profile webhooks/API once decode stops dominating.

Full data: `protobench/RESULTS.md`. Conversion spec to follow on this branch.

Issue #42, opened 2023-01-25
We need to clear pokestop cache before transition?


Fabio1988:
https://github.com/UnownHash/Golbat/blob/cd93a82fd0f7a8bbddc8c6e4327b96a961c47e66/stats.go#L237

if pokestop transitioned to gym this cache entry should expire if not raw processed anymore, so imo we could keep the whole cache in that case?! :)

jfberry:
Agreed, I did cut this out but it's crept back in with a merge

Fabio1988:
I remembered that too, was confused now 🤣

Fabio1988:
[65760482d380bc6464e291a00f67a219f4b0b96a](https://github.com/UnownHash/Golbat/commit/65760482d380bc6464e291a00f67a219f4b0b96a)

PR #381, opened 2026-07-07
hyperpb decode engine: GMO/Encounter/DiskEncounter on arena-based lazy parsing

jfberry:
## What this is

GMO, Encounter, and DiskEncounter decoding moved from protobuf-go struct unmarshal to **hyperpb** (buf.build/go/hyperpb v0.1.3: compiled parser tables, arena allocation with reuse, runtime profile-guided recompilation) behind **`pogoshim`** — a generated typed-accessor layer that works identically over hyperpb messages and std protobuf-go structs. Measured basis (protobench harness, frozen prod corpus, PR #378): +70% GMO decode rate, −78% bytes and −81% objects per decode, GC CPU share 5.1%→1.4%; **3.2–3.6× decode throughput under a prod-shaped 2 GB live heap**.

Base is `perf/eviction-lock-contention` since that's what the target deployment runs.

## Architecture (one accessor surface, two engines)

- `cmd/pogoshimgen` generates `pogoshim/` (165 messages): protoc-style getters backed by cached FieldDescriptors. **Strings/bytes are cloned out of the arena** (hyperpb is zero-copy; retaining a view would pin an entire payload copy — regression-tested in `pogoshim/retention_test.go`). Wrapping nil/invalid messages yields zero shims that chain safely.
- `protoengine.go` + `protoengine_hyperpb.go` (build-tagged amd64/arm64, std-forcing stub elsewhere): per-method engine selection, `hyperpb.Shared` arena pools freed exactly when the decode closure returns, PGO warmup over the first 256 packets/method then a one-time recompile.
- The std engine decodes with protobuf-go and wraps `ProtoReflect()` into the same shims — every decoder code path is shared, so **rollback is config, not code**.
- Decoder call graph (fort/station/incident, pokemon wild/nearby/map, weather, spawnpoint) migrated to shim signatures; retention boundaries eliminated: weather consensus now stores value structs, the disk-encounter cache stores raw payload bytes decoded on consumption.

## Config (defaults shown; see config.toml.example)

```toml
[proto_engine]
gmo = "hyperpb"              # "std" = instant rollback, no rebuild
encounter = "hyperpb"
disk_encounter = "hyperpb"
shadow_sample_rate = 0.01    # 1% of packets dual-decoded and digest-compared
pgo = true
```

## Safety nets

- **Shadow verification**: sampled packets decode through BOTH engines; field-level digests must match. Divergence → `golbat_proto_shadow_total{result="mismatch"}` + ERROR log. Expect zero mismatches; watch this and RSS during rollout.
- Cross-engine differential tests (GMO walk, encounter chain, consensus semantics, presence-only divergence) + race tests in CI scope.
- Known behavioral change (deliberate): malformed GMO bytes now return a parse error instead of the old fall-through that processed a zero-valued proto and reported a fabricated non-success status.

## Review pointers

- `pogoshim/pogoshim.gen.go` is generated — spot-check, don't line-review. The generator (`cmd/pogoshimgen/main.go`) is the thing to read.
- The interesting files: `protoengine*.go`, `protoshadow.go`, `decode.go` (GMO closure), `decoder/weather_consensus.go` (value-struct retention), `decoder/main.go` (Raw carriers + disk-encounter bytes cache).
- Review history (per-task spec+quality gates + whole-branch final review) caught and fixed: a PGO warmup data race, an arena use-after-free in the ProactiveIVSwitch goroutine capture, shadow-digest coverage holes, and the arena string-pinning issue above.

## Wave 3 (not in this PR)

The remaining ~15 low-volume methods (fort details, gym info, quests, routes, tappables, GET_MAP_FORTS) stay on std. Note `getMapFortsCache` retains decoded protos — GET_MAP_FORTS must keep the std engine (or copy values) until its own wave.


jfberry:
## Wave 3 pushed: every method now on the engine (12 new commits)

All remaining client-proto methods migrated to the hyperpb engine behind pogoshim: fort details, gym info, quests (the 280-line reward/condition switch), map forts, routes, incidents/invasions, nebula battle state, tappables, stations, contests, RSVPs, and social/player. The only exception is the push-gateway path (reads two scalars, nothing crosses the decoder boundary — documented in code).

**Foundation changes:**
- Config scales via `proto_engine.default = "hyperpb"` + `[proto_engine.overrides]` per-method pins; your existing `gmo`/`encounter`/`disk_encounter` keys still win if set. Unknown values warn at startup and run std.
- Shadow verification covers every migrated method via a generic descriptor-ordered digest (single + request/data composite entries) — same `golbat_proto_shadow_total` metric.
- Last proto-retention site eliminated: `getMapFortsCache` now stores value structs.
- Generator hardening along the way (each found by review or implementation, each regression-tested): cloning `StringList` for repeated strings (second incarnation of the arena-pinning bug — repeated fields bypassed the singular-getter clone), `Has*()` for scalar oneof members (tappable rewards would have silently miscategorized), and a faithful GoCamelCase port for nested type names.
- New guide: `docs/adding-a-proto.md` — the seven-step path for wiring any future proto through generator → engine handle → decode entry → shadow → config.

**Rollout notes for the target deployment:**
- Startup now compiles 28 hyperpb message types and logs `[PROTO_ENGINE]` per method; PGO warmup double-parses the first 256 packets per method (or gives up after 10 min for rare methods) — one-time cost.
- Watch `golbat_proto_shadow_total{result="mismatch"}`: note a sampled *malformed* payload counts as mismatch too (both engines fail identically); the `[PROTO_SHADOW] decode failed` log line disambiguates from a real digest divergence.
- Rollback remains per-method config: `[proto_engine.overrides] quest = "std"` etc., or `default = "std"` for everything.

Deferred fast-follows (from the wave-end review, none blocking): empty-ID guards for pathological SUCCESS payloads missing their primary submessage (previously a process-crashing nil-deref, now a degraded row), generator map-field support if a second consumer appears, two stale comment references.

jfberry:
## Shadow verification catches its first real bug: hyperpb `Recompile` corrupts repeated strings

Live `fort_details` traffic on the test system produced `[PROTO_SHADOW] digest mismatch` errors within seconds of deploy at 1% sampling. Diagnosis trail:

1. The mismatching payload decodes **identically** under std and baseline-hyperpb on a dev machine — not payload-deterministic.
2. hyperpb passed every conformance probe (explicit-zero presence, duplicate-field last-wins/merge, non-minimal varints).
3. The variable was **runtime PGO**: after warming a profile and calling `ty.Recompile(profile)`, the recompiled parser returns `image_url` (repeated string) with **2 elements instead of 1**. Reproduced with a fully synthetic payload — upstream bug in hyperpb v0.1.3.

**Mitigation (commit 8034dbb):** `proto_engine.pgo` now defaults to **false** — the baseline compiled parser is unaffected and passes all equivalence gates; this costs only the measured ~4% recompile bonus. A canary test (`TestHyperpbRecompileRepeatedStringDuplication`) asserts the bug is present and deliberately fails when a hyperpb upgrade fixes it — the signal to re-enable the default. Go compiler PGO (`default.pgo`) is unrelated and stays on.

Impact window on the test system: fort image URLs read `At(0)`, and the duplicated element is a copy — so no corrupted values landed; the mismatch counter did exactly what it was designed to do.

Suggested upstream issue (bufbuild/hyperpb-go), happy for anyone to file:

> **Title:** Profile-guided Recompile duplicates repeated-string field elements
> **Body:** Compile a MessageType from a descriptor with a `repeated string` field; parse a payload containing one element (correct, len=1). Record a profile over ~300 parses of that payload via `WithRecordProfile(profile, 1.0)`, then `ty.Recompile(profile)`. Parsing the same payload with the recompiled type yields len=2 (element duplicated). Reproduces on arm64 (darwin) and amd64 (linux), hyperpb v0.1.3, protobuf-go v1.36.x. Minimal repro: [canary test in our repo] — happy to provide a standalone reproducer.

jfberry:
## Runtime PGO re-enabled — upstream fix verified against our protos

hyperpb fixed the repeated-string duplication bug (issue #39 → PR #40, merged 2026-07-13). Two more correctness fixes landed the same day, both worth having: **#41** (astral-plane UTF-8 validation) and **#42** (data loss on 64-bit varint oneofs). No tagged release yet, so pinned to the main tip carrying all three: `v0.1.4-0.20260713222322-15810fedcdbe`.

**Verified against our own protos, not just the synthetic repro:** warmed a PGO profile, recompiled the `fort_details` parser, then shadow-compared it against std on the **exact production payload that originally failed** plus all 200 `FORT_DETAILS` corpus samples — all 201 now agree, and `image_url` decodes with the correct list length under the recompiled parser.

Changes:
- `proto_engine.pgo` defaults back to **true** (reclaims the ~4% recompile win)
- the canary flipped from "assert bug present" to a permanent **regression guard** — it now asserts the recompiled parser does *not* duplicate, so a hyperpb downgrade/regression breaks CI instead of silently corrupting data
- config example + CLAUDE.md updated

Full suite + race green, `go vet` clean. Shadow verification stays on at 1% as the permanent tripwire.

jfberry:
Conclusion: whilst there are some GC gains, and marginal cpu improvement this does not give payback given the maintenance burden

PR #382, opened 2026-07-11
feat: parse temp evolution branch quest rewards

Mygod:
## Summary

- update the bundled generated protobuf definitions to Base 0.417.x using the exact `pogo/vbase.pb.go` from #360
- decode `TEMP_EVO_BRANCH_RESOURCE` (quest reward type 20) into `amount`, `pokemon_id`, and optional `temp_evolution` fields
- populate the indexed first-reward amount and Pokemon columns
- document the new reward payload and cover it with a protobuf wire round-trip regression test
- carry over #360's generated-protobuf lint exclusion and timeout increase, which are required for the larger generated file to pass CI

## Why

The protobuf bundled on `main` does not define reward type 20 or its field-24 payload. Golbat therefore preserves the numeric reward type but emits an empty `info` object and cannot populate the indexed reward details.

This keeps the upstream reward type truthful rather than rewriting type 20 as the older generic mega-resource type 12.

## ReactMap coordination

[WatWowMap/ReactMap@8756f38f](https://github.com/WatWowMap/ReactMap/commit/8756f38f964c7a330e16fb57888795e5bbd5da16) currently normalizes only the old empty type-20 payload through a temporary Mewtwo fallback. Once Golbat emits populated type-20 details, ReactMap needs a companion change that consumes those details as mega energy while retaining the empty-payload fallback for older Golbat deployments.

Keep this PR in draft until that consumer-side handling is ready if ReactMap compatibility is required for deployment.

## Validation

- `go test ./...`
- `golangci-lint v2.12.2 run` - 0 issues
- generated protobuf blob verified byte-for-byte against #360

### DIFF
--- a/decoder/pokestop_decode.go
+++ b/decoder/pokestop_decode.go
@@ -343,6 +343,24 @@ func (stop *Pokestop) updatePokestopFromQuestProto(questProto *pogo.FortSearchOu
 				rewardAmount = null.IntFrom(int64(info.Amount))
 				rewardPokemonId = null.IntFrom(int64(info.PokemonId))
 			}
+		case pogo.QuestRewardProto_TEMP_EVO_BRANCH_RESOURCE:
+			info := rewardData.GetTempEvoResource()
+			if info == nil {
+				break
+			}
+			infoData["amount"] = info.Amount
+			if isFirst {
+				rewardAmount = null.IntFrom(int64(info.Amount))
+			}
+			if branch := info.GetTempEvoPokemonBranch(); branch != nil {
+				infoData["pokemon_id"] = int(branch.PokedexId)
+				if tempEvolution := int(branch.TempEvoId); tempEvolution != 0 {
+					infoData["temp_evolution"] = tempEvolution
+				}
+				if isFirst {
+					rewardPokemonId = null.IntFrom(int64(branch.PokedexId))
+				}
+			}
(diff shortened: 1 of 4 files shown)


decoder/pokestop_decode.go
```diff
@@ -343,6 +343,24 @@ func (stop *Pokestop) updatePokestopFromQuestProto(questProto *pogo.FortSearchOu
...
+				break
+			}
+			infoData["amount"] = info.Amount
+			if isFirst {
+				rewardAmount = null.IntFrom(int64(info.Amount))
+			}
+			if branch := info.GetTempEvoPokemonBranch(); branch != nil {
+				infoData["pokemon_id"] = int(branch.PokedexId)
+				if tempEvolution := int(branch.TempEvoId); tempEvolution != 0 {
+					infoData["temp_evolution"] = tempEvolution
```
ccev, line 358: Naming is a bit inconsistent across Golbat, but I'd prefer `temp_evolution_id` as the name (i don't care too much tho)

ccev:
> Keep this PR in draft until that consumer-side handling is ready if ReactMap compatibility is required for deployment.

ReactMap depends on Golbat, not the other way around. This PR does not have to be a draft. I'm running this now

Mygod:
Don't read too much into that. It's GPT written. :joy:

PR #387, opened 2026-07-26
feat: optionally break down shiny stats by geofence area

pjorritsma:
## Summary
- `pokemon_shiny_stats` was always written as a single world-only row per pokemon/form — inconsistent with `pokemon_stats`, `pokemon_hundo_stats`, `pokemon_nundo_stats`, `raid_stats`, `invasion_stats`, and `quest_stats`, which all break down by geofence area (with an "unmatched" fallback).
- Adds a `shiny_stats_per_area` config flag (default `false`) under `[stats_intervals]`. When enabled, shiny stats use the same area/fence lookup and unmatched/world fallback pattern already used by `updatePokemonStats` and the other stats functions.
- No schema changes needed — `pokemon_shiny_stats` already has `area`/`fence` columns.

## Behavior
- Flag disabled (default): unchanged — one row per pokemon/form under `world`.
- Flag enabled: one row per pokemon/form per matched geofence area (plus `unmatched`/`world` fallback), same write pattern as `pokemon_hundo_stats`/`pokemon_nundo_stats`. This increases write volume proportionally to the number of configured geofences, so it's opt-in.

## Test plan
- [ ] `go build ./...`
- [ ] Manually verify with `shiny_stats_per_area = false`: shiny stats still land only under `world` as before.
- [ ] Manually verify with `shiny_stats_per_area = true` and a geofence configured: shiny stats land under the matching area/fence plus `world`.

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -121,3 +121,4 @@ proactive_iv_switching_to_db = false # Write proactive IV changes to database (d
 #raid_stats_interval_minutes = 10       # Interval for writing raid stats (default: 10 minutes)
 #invasion_stats_interval_minutes = 15   # Interval for writing invasion stats (default: 15 minutes)
 #quest_stats_interval_minutes = 15      # Interval for writing quest stats (default: 15 minutes)
+#shiny_stats_per_area = false           # Break down pokemon_shiny_stats by geofence area instead of world-only (default: false). Increases write volume similar to pokemon_hundo_stats/pokemon_nundo_stats.
--- a/config/config.go
+++ b/config/config.go
@@ -169,11 +169,12 @@ type weather struct {
 }
 
 type statsIntervals struct {
-	PokemonStatsIntervalMinutes  int `koanf:"pokemon_stats_interval_minutes"`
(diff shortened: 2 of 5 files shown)


jfberry:
Shiny stats data is to calculate the odds which are global
I haven’t looked at the implementation but I don’t see the value in this.

pjorritsma:
Fair point, and I get the concern — but odds are still fully global here, nothing changes on that front. I kept that in mind while making this change.

Looking at the code, the world row gets written unconditionally regardless of the flag (stats.go always appends {Parent: "world", Name: "world"} before writing), so whatever calculates odds off that row keeps working exactly as before.

What the flag adds is just extra rows per matched geofence on top of that world row — useful if you want to see shiny counts per area, e.g. to compare scan coverage or spot regional event boosts. It's opt-in and defaults to false, so unless someone flips it on, behavior and write volume stay identical to today. And even with it on, the global world numbers aren't touched at all.

Happy to add more detail or rework the approach if there's still concern about the write volume.

jfberry:
Shiny counts are meaningless since they are per user. They only have meaning as odds. Which are global. 
so this doesn’t give you anything new

PR #391, opened 2026-07-29
Request-first lure pokemon processing

### LINKED ISSUE #283: Disk (lure) encounter cache no longer needed
The disk cache was put in place because the GMO was needed to find the fort id (and lat/lon of the pokemon).
This could be avoided entirely by processing the request.
In fact, it probably means that all GMO processing of lures and the whole disk cache could be removed

```protobuf
message DiskEncounterOutProto {
	enum Result {
		UNKNOWN = 0;
		SUCCESS = 1;
		NOT_AVAILABLE = 2;
		NOT_IN_RANGE = 3;
		ENCOUNTER_ALREADY_FINISHED = 4;
		POKEMON_INVENTORY_FULL = 5;
	}

	Result result = 1;
	PokemonProto pokemon = 2;
	CaptureProbabilityProto capture_probability = 3;
	Item active_item = 4;
	int32 arplus_attempts_until_flee = 5;
}

message DiskEncounterProto {
	int64 encounter_id = 1;
	string fort_id = 2;
	double player_lat_degrees = 3;
	double player_lng_degrees = 4;
	double gym_lat_degrees = 5;
	double gym_lng_degrees = 6;
	Item disk_item_id = 7;
}
```

### LINKED ISSUE #390: Lure Pokemon becomes unqueryable after Pokestop lookup failure
## Summary

When Golbat processes a lure `MapPokemonProto`, a failed Pokestop lookup leaves a partially initialized Pokemon record. The caller still saves that record, making it unrepairable for the remainder of its cache lifetime.

This can produce the following contradictory symptoms:

- A successful `DiskEncounter` adds the species and IVs and emits a `PokemonIV` webhook.
- The webhook can therefore produce a lure hundo alert.
- `/api/pokemon/v2/scan` never returns that Pokemon around the lure.
- Replaying the GMO after the Pokestop becomes available does not repair it.

This was reproduced on `main` at `200c6d9653606494df58c7ff9f0b61087de95088` with `pokemon_memory_only = true`.

## Code path

`updateFromMap` returns when `mapPokemon.SpawnpointId` cannot be resolved:

https://github.com/UnownHash/Golbat/blob/200c6d9653606494df58c7ff9f0b61087de95088/decoder/pokemon_decode.go#L180-L197

The lookup error is discarded, so a missing Pokestop and a database error follow the same path.

`UpdatePokemonBatch` nevertheless applies any cached DiskEncounter and unconditionally saves the record:

https://github.com/UnownHash/Golbat/blob/200c6d9653606494df58c7ff9f0b61087de95088/decoder/gmo_decode.go#L179-L195

Saving the new record:

- marks it as no longer new;
- inserts it into the R-tree at its zero-value coordinates;
- leaves `pokestop_id` and `expire_timestamp` unset.

https://github.com/UnownHash/Golbat/blob/200c6d9653606494df58c7ff9f0b61087de95088/decoder/pokemon_state.go#L250-L292

A later DiskEncounter supplies the species, IVs and `lure_encounter` seen type, but not the location or expiry:

https://github.com/UnownHash/Golbat/blob/200c6d9653606494df58c7ff9f0b61087de95088/decoder/pokemon_decode.go#L820-L825

Finally, the scan response drops records without a future expiry:

https://github.com/UnownHash/Golbat/blob/200c6d9653606494df58c7ff9f0b61087de95088/decoder/api_pokemon_response.go#L187-L201

A later GMO cannot repair the record because `updateFromMap` immediately returns for every non-new record.

## Reproduction

A temporary package-level regression test performed the following:

1. Enabled `pokemon_memory_only`.
2. Processed a lure `MapPokemonProto` whose `SpawnpointId` was absent from both the Pokestop cache and `GeneralDb`.
3. Confirmed the cached Pokemon had:
   - `newRecord = false`
   - `lat = 0`
   - `lon = 0`
   - no `pokestop_id`
   - no `expire_timestamp`
4. Processed a successful 15/15/15 `DiskEncounterOutProto`.
5. Confirmed a `PokemonIV` webhook was emitted with:
   - `latitude = 0`
   - `longitude = 0`
   - `disappear_time = 0`
6. Added the missing Pokestop and replayed the original GMO.
7. Confirmed the record remained at `0,0` without an expiry.
8. Confirmed:
   - the internal R-tree contained the encounter at `0,0`;
   - the clean v2 API omitted it because its expiry was unset;
   - a scan around the real Pokestop returned no candidate.

The regression test passed consistently.

## Expected behavior

A failed base-location update should not commit the Pokemon as a completed, non-new record.

At minimum:

- a missing Pokestop or lookup error must not permanently poison the encounter;
- a later valid GMO must be able to populate the location and expiry;
- an incomplete record must not be indexed at `0,0`;
- lookup errors should remain observable instead of being discarded.

Using the fort ID and coordinates from the DiskEncounter request may also remove the dependency on a prior GMO. That is already discussed in #283:

https://github.com/UnownHash/Golbat/issues/283

However, the save/repair invariant should remain safe even if the DiskEncounter request is not processed.

jfberry:
Fixes #390. Closes #283.

## Problem

A failed pokestop lookup while processing a lure `MapPokemonProto` committed a partially initialized pokemon record: indexed in the R-tree at (0,0), no `pokestop_id`, no expiry, marked non-new — and therefore permanently unrepairable, while a later DiskEncounter would happily attach IVs and fire hundo webhooks at lat/lon 0 with `disappear_time` 0 (#390). Structurally, the whole `diskEncounterCache` existed only because the DiskEncounter *response* carries no location, forcing responses to wait for a GMO (#283).

## Approach

Each proto contributes what it alone knows, in any arrival order:

| Source | Contributes |
|--------|-------------|
| GMO fort entry (`fort.ActivePokemon`) | placement (fort id + coords, captured at extraction) and the verified despawn time |
| `DiskEncounterProto` (request) | placement for encounter-first arrivals (encounter id, fort id, fort coords) |
| `DiskEncounterOutProto` (response) | species, IVs, CP, level |

- **No lure path touches the pokestop record anymore.** The lure pokemon rides inside its fort's own GMO entry, so `RawMapPokemonData` now carries the fort's id/lat/lon captured at extraction — placement cannot fail, even in pokemon-only scan contexts where the fort cache is never populated.
- **`decodeDiskEncounter` now requires and parses the request proto** (same posture as `GetStationDetails`): encounter id comes from `request.EncounterId` (the `PokemonDisplay.DisplayId` stand-in is gone), and an encounter with no prior GMO creates a fully placed `lure_encounter` record with a 180 s unverified expiry (lure spawns live 3 minutes). A later GMO tightens it to the verified despawn. Requests missing the payload are counted (`request_missing`) and skipped.
- **`updateFromMap` is now an order-tolerant merge**: new records get placed from the captured fort fields; existing lure records only receive GMO-owned facts (verified expiry, cell id, username backfill) — never a seen-type downgrade, never touching encounter data, and no-change sightings skip the save.
- **`diskEncounterCache` is deleted** — nothing waits for anything.

## Testing

- Regression tests mirror the #390 repro end-to-end: unknown-fort GMO placement, encounter-first creation (webhook carries real coords and a future expiry), GMO-after-encounter merge, GMO-first upgrade, request-missing skip, and v2/v3 scan visibility of estimated-expiry records.
- `go build`, `go vet`, full `go test ./...`, and `golangci-lint run` are clean; the new tests pass under `-race`.
- Note: a pre-existing (on `main`) test race in `decoder/station_battle_test.go` (direct `statsCollector` reassignment vs the async stats worker) was found while verifying — left for a follow-up since it's unrelated to this branch.


Mygod:
Thanks — this looks like it fixes #390 and substantially improves the request-first `DiskEncounter` path. However, I found a separate live-wire issue that this PR does not currently cover.

I captured and decoded current `GET_MAP_OBJECTS` traffic:

- 17,009 GMO payloads
- 67 unique lure Pokémon in `fort.ActiveFortPokemon`
- zero occurrences of singular `fort.ActivePokemon`
- 65 of those lure IDs later received a successful `DiskEncounter`
- none appeared through Nearby, Wild, Catchable, Tappable, or normal Encounter paths

At `aebc868`, `decodeGMO` still extracts only singular `fort.ActivePokemon`:

https://github.com/UnownHash/Golbat/blob/aebc8681b48004692bef44701ee7bcc241f2c8bc/decode.go#L553-L561

Consequently, PR #391 should make Disk-encountered lure Pokémon visible using the request coordinates, but:

- lure Pokémon without a `DiskEncounter` remain absent;
- repeated-only GMOs cannot provide the verified `ExpirationTimeMs`;
- the new GMO placement/merge logic is never reached for the current wire format.

There being only one active lure Pokémon per stop is compatible with this: the repeated field simply contains one `LURE` wrapper while the singular field remains nil.

I think the repeated-field ingestion should be tracked separately from #390. The eventual fix should retain singular support, iterate `ActiveFortPokemon`, unwrap non-nil `PokemonProto` entries, filter `SpawnType_LURE`, and deduplicate by encounter ID if both representations appear.

A `decodeGMO` regression test with `ActivePokemon == nil` and one repeated `ActiveFortPokemon` entry would catch this. The current tests construct `RawMapPokemonData` directly and therefore bypass the missing extraction step.

jfberry:
Confirmed and fixed in 23a2334 — thanks for the capture work, and to the fixture for making this trivially testable.

Root cause matched your analysis exactly: extraction only read the singular `fort.ActivePokemon`, which is nil on current wire traffic; lure pokemon arrive as repeated `fort.ActiveFortPokemon` wrappers (`SpawnType == LURE`, nested `MapPokemonProto`). Notably the nested proto's lat/lon are zero on the wire, so the enclosing fort's coordinates are the only usable ones — the capture-at-extraction design this PR introduced already supplies them, so the fix stayed contained to the extraction layer.

What landed:
- `extractFortMapPokemon` unwraps LURE entries from the repeated field, retains legacy singular support, dedupes by encounter ID when both representations appear, and filters `POWER_UP` wrappers (nil wrapper/nil inner proto safe).
- Extraction-layer regression tests driven by a sanitized live-capture fixture (`testdata/gmo-active-fort-pokemon.pb`) with `ActivePokemon == nil` and one repeated LURE entry — addressing the gap you pointed out where the existing tests constructed `RawMapPokemonData` directly.

Folded into this PR rather than tracked separately since without it the PR's GMO merge path is dead code on current traffic.

PR #393, opened 2026-08-03
feat(api): status endpoint, fort scan limit status, availability enrichment

jfberry:
Golbat-side prerequisites for the Diadem fort API migration (ccev/diadem#174 — the consumer design this serves).

**Includes #392 as a cherry-pick with attribution preserved (thanks @ccev) — closes #392.**

## Deliverables

1. **`GET /api/status`** — secret-gated but **not** `fort_in_memory`-gated: reports enabled optional features and the server's scan caps so consumers can detect capabilities (and clamp their request limits) without probing gated endpoints:

   ```json
   {
     "features": { "fort_in_memory": true },
     "limits": { "max_pokemon_results": 3000, "max_fort_results": 4000 }
   }
   ```

2. **`limit_reached` on the fort scan responses** (gym/pokestop/station/combined) — straight mirror of #392: computed from the pre-collection key count against the effective limit (`fortScanLimit` = request limit clamped by `tuning.max_fort_results`).

3. **Availability enrichment** — `temp_evolution_id` on gym raid availability (mega/primal bosses were indistinguishable from base forms) and `ranking_standard` on showcase availability. Distinct evolutions now yield distinct entries; both fields thread through `FortLookup` → availability keys → readers.

4. **Fix: pokestop `first_seen_timestamp` was `int16`** on both the model and `ApiPokestopResult` — a unix timestamp wraps int16, truncating DB reads and API output (gym's field was already int64). Note while fixing: nothing in Golbat currently *assigns* this field for pokestops (existing DB values round-trip; new stops insert 0) — populating it on first insert is a separate follow-up.

## Verification

- `go build -tags go_json ./...`
- `go test -tags go_json ./decoder/ .` (new: fort limit helper tests, `/api/status` route tests incl. 401 + not-503-gated assertions, availability enrichment tests incl. mega-vs-base distinctness)
- `go test -tags go_json -race -run 'Availability|Observe…' ./decoder/`
- `golangci-lint run` — 0 issues


jfberry:
Memory note for the enrichment: the two new FortLookup fields are deliberately `int8` placed into existing alignment padding (RaidLevel's pad byte, and the pad before `ShowcaseExpiry`). Measured with `unsafe.Sizeof`: FortLookup was 184 bytes before this PR, 192 with naive `int16` placement, and **184 again as committed** — the enrichment adds zero bytes per resident fort. No DNF filter fields were added; the values feed only the availability index keys (mega X/Y distinctness in pick lists — raid level alone can't distinguish those).

jfberry:
Added the native-JSON completion for the remaining fort blob columns (`stationed_pokemon`, `showcase_focus`, `showcase_rankings`, both `quest_conditions`) via `jsonRaw` — the #385 design said native JSON for blobs but the pass only fully covered gyms. Wire-breaking for those five fields, but both consumers are unmerged: ReactMap's branch already handles both forms, and ccev/diadem#175 now does too (dual-mode `blobToString`).

PR #395, opened 2026-08-17
perf: intern pokestop_id and username into 4-byte handles

TurtIeSocks:
Split out of #394 at @jfberry's request. Stacked on `c/golbat-memory-persistence-6846cc` — the diff below is the five interning commits only.

Interns `Pokemon`'s `pokestop_id` and `username` into 4-byte handles. On top of #394's packing, this takes `Pokemon` from 352 to **312 bytes**, moving the allocator size class 352 → **320**, and removes roughly **3.25M heap objects** of duplicated fort ids.

## Why

`pokestop_id` was a `null.String` holding a ~35-character fort id, allocated per pokemon. At the 3.25M cached measured in production that is 3.25M heap objects and ~114 MB of duplicates against maybe 500k distinct forts. Bytes are the smaller half of that: GC mark cost tracks objects and pointers, and the same instance carries 34.6M live objects.

| | | |
|---|---|---|
| `PokemonData` | 256 B | **216 B** |
| `Pokemon` | 312 B | allocator class 352 → **320** |
| at 3.25M cached | | ~104 MB from the class move, ~114 MB of duplicate strings, ~3.25M objects |

## How

A `uint32` handle into a concurrent append-only table, with `sql.Scanner`, `driver.Valuer` and `MarshalJSON` so the database and JSON boundaries are unchanged. A phantom type parameter makes a wrong-table lookup a compile error at zero byte cost.

Publication ordering matters and is deliberate: the reverse slice is stored *before* the forward map entry that hands out the handle, so no window exists where a live handle resolves to nothing. Copy-on-grow is safe for stale readers because published slice headers are never mutated and entries never move.

Strings are capped at their column widths at intern time. The cap counts **characters, not bytes** — the columns are utf8mb4, so a byte cap would reject a legitimate 30-character multi-byte username — and a coarse `len` bound runs *before* the map probe, so an oversized input is refused without being hashed. An over-length string is **rejected rather than truncated**: a clipped fort id is a plausible-looking key that `createPokemonWebhooks` would hand to a pokestop lookup, whereas a null handle is true and self-heals on the next sighting. A rejected id falls to the same `"None"` path a pokemon with no nearby pokestop already uses.

An unresolvable handle degrades to `nil` at every boundary including `Value()`. That last one is deliberate: the write-behind flush treats a `driver.Valuer` error as non-retryable and discards the whole batch — 50 rows normally, 1000 during a shutdown preserve — and it would repeat every save, since the handle lives in the cached entity. The failure is already counted and logged inside `resolve()`, so erroring bought no visibility while costing unrelated rows.

## Open questions @jfberry raised, not yet actioned

**Username may not need interning at all.** Its only load-bearing consumer is the shiny/duplicate-encounter dedup in `updateEncounterStats`, which needs an account-distinguishing token transiently at encounter time — and the decode context already has the username in scope. The persisted field is set once and otherwise only flows to the column and the API response. Config-gating whether it is persisted would make the harder half of this problem — a caller-supplied, optionally-unauthenticated, unbounded key set retained forever — disappear rather than be engineered around.

**If it stays, the count needs bounding by construction.** The file's own principle is "bounded by construction rather than by hope", and that is currently applied to string length but not to table size. Username arrives off the request body with the `RawBearer` gate optional. Under the old representation a junk username was transient, evicted with the entity in about an hour; interned, it is resident until restart, at request rate. Degrading to the null handle past a ceiling would match every other failure path here.

**Pokestop ids are already integers.** 32 hex characters plus `.` and a short type suffix — a hex rendering of a 128-bit identifier, which is why the column is `varchar(35)`. So a fixed-width value type (`[16]byte` + suffix) converted at the boundaries, the way `Uint64Str` already handles encounter ids, may beat a shared table: no global state, trivially comparable and hashable. It costs ~13 more bytes per pokemon than the handle and would push 312 back into the 352 class, so for the cached `Pokemon` alone the handle is tighter — but the payoff is the fort subsystem, where cache keys, `FortLookup` keys and R-tree entries all become fixed-size value keys with no 35-byte string hashing and no GC-visible key pointers. #384 parked the incident-id version and documents the real risk: parse-on-ingest must be fallible with a graceful fallback, so a nonconforming id degrades rather than disappears.

Both shapes, or a combination, are worth evaluating here before committing to the table as the foundation. Marked draft for that reason.

## Testing

- Three MariaDB round-trip tests against a real instance, including one proving rows written before this change still load. `TestPokemonFullRowRoundTrip` reads `pokestop_id` back as a raw string — without a recognised `driver.Valuer`, `database/sql` would have written the integer handle into a varchar column, and a self-consistent Scanner/Valuer pair could never detect it.
- `TestInternIsCanonical` and `TestInternTableConcurrent` pin the canonicality that the setters' `!=` dirty check depends on.
- `TestInternRejectsOverlongStrings` covers the wide-character boundary both ways and a 1 MiB string that never reaches the map.
- `TestInternProductionTablesMatchColumnWidths` ties the caps to the migration files.
- `golbat_intern_table_size`, `golbat_intern_lookup_failures_total` and `golbat_intern_rejected_total` are exported, the last kept distinct so "always a bug" keeps its meaning.
- Green under both build tags, `-race`, and `golangci-lint`.

## Not measured

The production figures in #394 were captured before this landed, so the improvement here is arithmetic on measured inputs rather than an observation. A capture on this branch would settle it.


jfberry:
Following up on the two open questions in the PR description — we benchmarked the fort-id representations against each other before committing to the intern table as the foundation. Results below, and the direction we've settled on.

## Setup

Throwaway harness pinned to Golbat's own `rtree v1.10.0` / `xsync v4.5.0`, Apple M3 Pro. Realistic ids (32 lowercase hex + suffix, measured suffix mix), 40-city clustered coordinates. Three representations: **string** (today), **uint32 handle** (this PR's model, extended to the fort subsystem), **`[16]byte` GUID + 1-byte suffix enum** (value type). The encoding round-trips every id losslessly, and with the enum numbered `none < .11 < .12 < .16 < .23`, byte comparison of the value type matches varchar ordering — so the fort tracker's keyset pagination and station-battle `ORDER BY` grouping keep working against in-memory sorts.

## Hot paths (2M forts; 500k shows the same ordering with smaller gaps)

| Path | string | uint32 handle | [16]byte+suffix |
|---|---|---|---|
| FortLookup load per scan candidate | 381 ns | **207 ns** | 253 ns |
| R-tree search (~650 hits) | 8.9 µs | 8.0 µs | **8.0 µs** |
| **Full scan inner loop** (search + load per hit) | 181 µs | **99 µs (1.8×)** | 162 µs (1.1×) |
| Tree churn (delete+insert pair) | 937 ns | 916 ns | 979 ns |
| Ingest: proto string → key | ~0 | 127 ns | 218 ns |
| Emit: key → string | ~0 | **14 ns** | 103 ns |

Ingest/emit conversions are noise against a GMO decode or a JSON marshal.

## Resident heap (2M forts fully populated: entity cache + FortLookup + R-tree + both tracker maps, plus 3.25M pokemon holding a `PokestopId` — distinct string copies in the string rep, as production does today)

| | string | handle | bin |
|---|---|---|---|
| Heap | 2584 MB | **2339 MB** | 2475 MB |
| Heap objects | 20.6M | 19.4M | **15.3M** |
| Forced-GC median | 177 ms | 128 ms | **126 ms** |
| Forced-GC p90 | 196 ms | 144 ms | **129 ms** |

The metric this whole arc is about — **GC mark time — is a tie: both options cut it ~28%** vs string. The handle is tightest on bytes (the table still holds all 2M strings, but pokemon stop duplicating them); the value type is tightest on object count (no id strings exist anywhere) and has the tightest GC tail. Synthetic heap — trust the relative deltas, not the absolute ms.

## Id format, verified against a production instance

A census across pokestop/gym/station found **every real fort id parses** into `[16]byte`+suffix:

- Suffixes observed: `.16`/`.11`/`.12` on pokestops and gyms, `.23`/`.16`/`.11`/`.12` on stations — the enum must be universal across fort types, and nothing outside that set exists.
- **Bare 32-char ids (no suffix) are a live Niantic id shape, not legacy junk**: they're sponsored/promotional forts (*EE*, *Community Ambassador Location*, *Play! Pokémon Store*, …), several updated within the hour of the census. `SuffixNone` is first-class, and it's also what makes byte order match varchar order.
- Exactly **one** nonconforming row exists (an empty-string pokestop id from an old ingest bug). Parse-on-ingest stays fallible with a graceful fallback, per the #384 risk — but the fallback is exercised only by junk. One design rule falls out: `Parse("")` must fail to the fallback, never map to the zero value, because zero *is* the "None" sentinel.

## Code-risk inventory (whole codebase)

- **Handle**: drops into every `comparable`/`cmp.Ordered` site unchanged. Costs: reverse-resolve at ~35 JSON/webhook/API emit sites; lookup-or-create entry points for arbitrary strings (`getFortIdFromContest`, API path params, `/api/*/query` id lists); a global append-only table that leaks deleted forts until restart.
- **Value type**: one blocking mechanical change — `writebehind.TypedQueue` constrains keys to `cmp.Ordered`, so it needs `comparable` + an explicit comparator (which preserves the deadlock-avoidance sort order byte-for-byte, since the encoding is order-preserving). No global state, no leak, emit formats locally.

## Way forward

**`[16]byte` + suffix enum, everywhere fort ids live — including `pokemon.PokestopId`.** On the metric that matters (GC) the two are equivalent, so the tiebreakers are structural: the value type deletes a category of machinery — global table, unbounded-growth ceilings, deleted-fort leak, resolve failure modes at 35 emit sites — rather than engineering around it, and drives heap objects to the floor. Its one real cost is `Pokemon` moving back to the 352 allocator class (~104 MB at 3.25M cached), which is size-golf we're deliberately not optimizing for. The handle's genuine win is scan throughput (1.8× vs 1.1×), but fort scans are low-traffic today; interning *on top of* value ids stays available as the pre-scoped lever if that changes.

Consequences for this PR:

1. **`pokestop_id` interning is superseded** by the value type — `interned_string.go` and its supporting machinery come out.
2. **Username stops being persisted by default.** New config option (default off); the account name is threaded from the decode context to the webhook payload and the shiny/duplicate-encounter dedup, which want the *current* account anyway. With the option on, the field is a plain `null.String` — with the value type `Pokemon` is in the 352 class regardless, so username interning no longer buys a size class and exits along with the table.
3. Next: design doc for the value type (encoding, fallback path, `TypedQueue` constraint change, boundary conversions), then the fort-subsystem conversion as stacked work.

Benchmark harness available on request — it was built as a spike and stays out of the tree; the parse/format/order-preserving encoding and its correctness tests carry into the real implementation.

🤖 Benchmarks and analysis by [Claude Code](https://claude.com/claude-code)

PR #397, opened 2026-08-19
feat(api): add Buddy Showcase focus filtering

Mygod:
## Summary

- add an exact Buddy Showcase `contest_focus` selector to Pokéstop and combined fort-scan DNF filters
- apply the structured focus predicate before `max_fort_results`, so unrelated Showcases cannot consume the endpoint cap
- expose active `showcase_focus` values as native JSON objects and advertise `showcase_focus_filter: true` through both availability endpoints
- load `showcase_focus` during cache-miss/startup population and keep only a compact Buddy-level projection in the hot fort lookup

Example filter:

```json
{
  "contest_focus": [
    {
      "type": "buddy",
      "min_level": 3
    }
  ]
}
```

## Compatibility

- existing `contest_pokemon` and `contest_pokemon_type` filters are unchanged
- existing availability `pokemon_id`, `form`, and `type_id` mirrors remain available
- legacy rows without structured focus return `showcase_focus: null`
- the existing per-stop `showcase_focus` wire representation is unchanged
- clients can treat a missing or false `showcase_focus_filter` capability as an older Golbat

Related to WatWowMap/ReactMap#1090.

## Verification

- `go test -tags go_json ./...`
- `go build -tags go_json ./...`
- focused `go test -race ./decoder/` coverage for parsing, indexing, DNF matching, cap ordering, and availability
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run` — 0 issues
- `git diff --check origin/main...HEAD`

### DIFF
--- a/decoder/api_fort.go
+++ b/decoder/api_fort.go
@@ -90,14 +90,23 @@ type ApiFortDnfFilter struct {
 	IncidentCharacter   []int16 `json:"incident_character" required:"false" doc:"Pokestop only: allowed incident character ids; omitted or null means no incident character constraint."`
 
 	// Pokestop - contest
-	ContestPokemon     []ApiDnfId `json:"contest_pokemon" required:"false" doc:"Pokestop only: allowed contest focus pokemon/form pairs; omitted or null means no contest pokemon constraint."`
-	ContestPokemonType []int8     `json:"contest_pokemon_type" required:"false" doc:"Pokestop only: allowed contest pokemon types; omitted or null means no contest type constraint."`
+	ContestPokemon     []ApiDnfId               `json:"contest_pokemon" required:"false" doc:"Pokestop only: allowed contest focus pokemon/form pairs; omitted or null means no contest pokemon constraint."`
+	ContestPokemonType []int8                   `json:"contest_pokemon_type" required:"false" doc:"Pokestop only: allowed contest pokemon types; omitted or null means no contest type constraint."`
+	ContestFocus       []ApiFortDnfContestFocus `json:"contest_focus" required:"false" doc:"Pokestop only: allowed structured contest focuses. Currently supports exact Buddy minimum levels. Omitted or null means no structured focus constraint; an empty list matches nothing."`
 
 	// Station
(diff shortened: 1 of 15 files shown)


jfberry:
Thanks for this — I've cherry-picked 8e10f6c into #393 (`feat/fort-api-status-and-limits`) as ea44536, with your authorship preserved, so the showcase filtering lands together with the availability enrichment it sits next to. #393 touches the same three structs, so keeping them on one branch avoids a merge race between the two PRs.

Conflicts were all additive and resolved by keeping both sides:

- `FortLookup` carries `ShowcaseBuddyMinLevel` alongside #393's `ShowcaseRankingStandard`. Both `int8`s share the existing pad before `ShowcaseExpiry`, so the struct is still 184 bytes (verified with `unsafe.Sizeof`) — your projection stays free.
- `showcaseKey` now dedupes on ranking standard as well as structured focus.
- `ApiPokestopShowcaseAvailable` returns both `ranking_standard` and `showcase_focus`.

Two small changes on top:

1. The `showcase_focus` nullability doc now reads "null for legacy rows without showcase_focus **and for rows whose stored focus could not be parsed**" — `updatePokestopLookup` logs and falls back to nil when `parseShowcaseFocus` fails, so a modern row with unparseable JSON also surfaces null.
2. `contest_ranking_standard` and `raid_temp_evolution_id` DNF filters were added, following the pattern your `contest_focus` established: your PR was the one that correctly wired a new dimension through lookup, availability index *and* filter in the same change, which is what made the two missing filters on the #393 side visible.

One thing worth flagging that this PR inherits rather than introduces. `ContestProto.Focuses` is a repeated field, and `updatePokestopFromGetContestDataOutProto` builds `focusStore` as a Go map, then calls `SetShowcaseFocus` once per iteration — so when a showcase carries more than one focus, only the last-visited entry survives, non-deterministically per decode (there's already a `SHOWCASE: we got more than one showcase focus` warn guarding that case). That was harmless while `showcase_focus` was informational, but it's now load-bearing for `ShowcaseBuddyMinLevel` and the `contest_focus` filter, so a dual-focus showcase could flicker in and out of matching between GMOs. Not a blocker for this change, and the fix belongs in the decode path — deriving the Buddy projection by scanning `focusStore` for a buddy entry specifically, rather than depending on whichever key iteration landed last.

Closing in favour of #393 once that merges — the commit and the credit carry over.

PR #400, opened 2026-08-20
Fix geofence queries: bind the fence, and test containment in Go

TurtIeSocks:
## Problem

The three geofence queries built their statement by concatenating the fence's JSON into it:

```go
"AND ST_CONTAINS(ST_GeomFromGeoJSON('"+string(bytes)+"', 2, 0), POINT(lon, lat))"
```

The fence arrives as a request body, and `geojson.Feature` round-trips its entire `properties` map through `MarshalJSON`. A property value is free-form text, so anything a caller puts there lands verbatim inside that quoted literal, including a quote of its own. `/api/quest-status`, `/api/pokestop-positions` and `/api/clear-quests` are all affected, along with the gin `NormaliseFenceRequest` path. All three require authentication, so this is not reachable anonymously.

## Approach

`FenceContainsPredicate` is now a constant carrying a bind placeholder, and `FenceQueryArgs` assembles the bounding-box corners and the fence JSON in the order every one of these queries binds them. The statement text no longer varies with the fence, so its contents have no literal to escape.

Which rows match does not change. The same predicates run in the same order against the same values; only the transport differs, with the fence passed as data rather than as text. Each query keeps its own comparison operators, including the `>`/`<` in quest-status against the `>=`/`<=` in the other two.

## The timeout, measured

I had assumed repeated GeoJSON parsing was the cost. That was wrong, and the measurement says so plainly. MariaDB 11.8, 1,000,000 pokestops with `ix_coords(lat, lon)`, a 2000-vertex fence (83 KB of GeoJSON) whose bounding box holds 628,560 candidates:

| Shape | Time |
|---|---|
| Current query, fence inlined | **55.0 s** |
| `SET @fence = ST_GeomFromGeoJSON(...)`, then `ST_CONTAINS(@fence, ...)` | **57.6 s** |
| The same through a CTE | **57.8 s** |
| `WITH ... AS MATERIALIZED` | syntax error on MariaDB |
| Bounding-box scan alone, all 628,560 rows read | 0.9 s |
| The same 628,560 points against a compiled fence in Go | 3.3 s |
| **Bounding box in SQL, containment in Go** | **≈ 4.2 s** |

Hoisting the parse out of the row loop changes nothing, and that is the result that matters: it rules out repeated parsing and leaves per-row `ST_CONTAINS` against a many-vertex polygon as the cost. Neither variant suggested in #398 moves the number, and one of them does not parse on MariaDB at all, since `AS MATERIALIZED` is PostgreSQL syntax.

So containment moves to Go. Candidates come from the bounding box, which the existing index already serves, and each one is tested against `geo.CompileFence` as the rows stream past, retaining only matches. That is the same matcher `MatchGeofences` already uses for stats and webhook area attribution, and it is differentially tested against orb.

Two points worth stating. A fence must be an area: a Polygon, a MultiPolygon, or a GeometryCollection holding polygons (flattened on the way in); a Point or LineString body is rejected with 400 rather than answered with an empty result, and the SQL containment path is gone entirely. And boundary semantics do not change: `CompileFence` counts the polygon boundary as inside, and so does MariaDB's `ST_CONTAINS` (checked on 11.8.2: a vertex point and an edge midpoint both return 1), so a fort exactly on a fence edge is attributed the same way before and after this change, and the same way stats and webhooks already attribute it.

These are synthetic numbers on uniform random data in a container, not a production table, so treat the ratio as the finding and not the absolute times.

One note for anyone revisiting the `SET @fence` form regardless: user variables are connection-scoped and the handle is pooled (`SetMaxOpenConns` in `main.go`). These call sites use `Select`, `Get` and `SelectContext`, which each take an arbitrary connection, so the `SET` and the `SELECT` can land on different ones. `@fence` would then be NULL, `ST_Contains(NULL, ...)` would be NULL, and every row would filter out: zero results, no error.

## Testing

Six tests in `db/geofence_query_test.go`. Three for the injection fix: the predicate keeps its placeholder and carries no literal to break out of, a hostile `properties` value survives only as a bound argument, and the bounding-box corner order the callers depend on. Three for the containment move: non-polygon fences fall back to SQL while polygons and multipolygons do not, the matcher's lat/lon argument order is right (the fences used have differing lat and lon spans, so a swap is visible), and boundary points count as inside.

`go build -tags go_json ./...` and `golangci-lint run` are clean, and the `db`, `decoder` and root suites all pass.

## Review follow-ups (pushed by maintainers)

- `CompileFence` no longer panics on a non-string `name`/`parent` property in a request body.
- The streaming queries record their outcome after the stream ends, so mid-stream failures are counted as errors.
- clear-quests stops at the first context error and the endpoint answers 500 instead of reporting a partial clear as ok.
- A Feature with a null geometry is rejected with 400 instead of a nil dereference.
- `GetPokestopPositions` and `GetQuestStatus` take the request context; all three queries bind one inclusive (`>=`/`<=`) bounding-box clause; dead `sql.ErrNoRows` checks removed.
- The SQL containment fallback is removed: non-area bodies get a 400 at the parse boundary, GeometryCollections of polygons are flattened onto the fast path, and `FenceContainsPredicate`/`FenceQueryArgs` are gone (the last MariaDB GIS use in the codebase).
- Tests: a capturing fake driver guards the emitted SQL end to end, distinct-corner bounding-box fixtures, and stream-error, cancelled-context, nil-geometry and non-string-property cases.

### DIFF
--- a/db/geofence_query.go
+++ b/db/geofence_query.go
@@ -1,0 +1,32 @@
+package db
+
+import "github.com/paulmach/orb/geojson"
+
+// FenceContainsPredicate is the SQL fragment testing that a row's (lon, lat)
+// lies inside a geofence.
+//
+// The fence is a bind parameter, never interpolated into the statement text.
+// It arrives as a request body, and a geojson.Feature round-trips its whole
+// properties map through MarshalJSON, so a property value containing a single
+// quote would otherwise close the string literal it was concatenated into.
+const FenceContainsPredicate = "ST_CONTAINS(ST_GeomFromGeoJSON(?, 2, 0), POINT(lon, lat))"
+
+// FenceQueryArgs returns the bind arguments for a geofence query: the bounding
+// box corners as min-lat, min-lon, max-lat, max-lon, then the fence JSON for
+// FenceContainsPredicate.
+//
+// Every geofence query places its bounding-box comparisons before the fence
+// predicate, so this order matches all of them. A query that departs from that
+// layout must not use this helper.
+func FenceQueryArgs(fence *geojson.Feature) ([]any, error) {
+	fenceJSON, err := fence.MarshalJSON()
+	if err != nil {
+		return nil, err
+	}
+	bbox := fence.Geometry.Bound()
+	return []any{
+		bbox.Min.Lat(), bbox.Min.Lon(),
+		bbox.Max.Lat(), bbox.Max.Lon(),
+		string(fenceJSON),
+	}, nil
+}
--- a/db/geofence_query.go
+++ b/db/geofence_query.go
@@ -1,6 +1,13 @@
 package db
 
-import "github.com/paulmach/orb/geojson"
+import (
+	"context"
+
+	"github.com/paulmach/orb"
+	"github.com/paulmach/orb/geojson"
+
+	"golbat/geo"
+)
 
 // FenceContainsPredicate is the SQL fragment testing that a row's (lon, lat)
 // lies inside a geofence.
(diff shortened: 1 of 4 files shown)


jfberry:
I don't think this actually moves the fence build out of the command (as per my message on the other issue).

TurtIeSocks:
Split out of #398 so it can land on its own. Two separate things in here.

First, the fence was being concatenated into the SQL text rather than bound. A geofence body round-trips its whole `properties` map through `MarshalJSON`, so a property value containing a quote ends up inside the string literal. All three endpoints require auth so it isn't reachable anonymously, but it is on main today.

Second, the timeout, with the numbers in the description. Short version is that hoisting the GeoJSON parse out of the row loop changes nothing, so the cost is per row `ST_CONTAINS` against a many vertex polygon. Bounding box in SQL plus containment in Go takes it from 55 s to roughly 4 s, and it helps every instance rather than only the `fort_in_memory` ones. Non polygon fences keep the SQL predicate so nothing exotic changes.

@jfberry does this shape look right to you? Happy to split the two commits into separate PRs if you'd rather take the injection fix on its own first.

jfberry:
This seems much more sensible; I'd have a couple of questions which are worth asking claude:
Elsewhere in the codebase we have used geofence optimisations (s2 cells, an optimised contains-in-geofence routine, and rtree map) - would any of these help here?

jfberry:
I'd also ask to be sure you are testing against mariadb (and which version) since mysql is not supported and has different geojson performance

jfberry (changes requested):
Thanks for this. It is the right change and I want it in. Splitting the injection fix from the timeout work, then measuring instead of assuming, is exactly how I'd want this handled. The `SET @fence` / CTE measurements settle the question I raised on #398, the pooled-connection note about user variables is a genuinely useful warning for anyone who revisits that idea, and moving containment onto `geo.CompileFence` reuses the matcher we already trust for stats and webhooks rather than inventing a second one. Thanks also for confirming the MariaDB version.

I verified it independently against a local MariaDB 11.8.2 with real pokestop rows, running the pre-PR `ST_CONTAINS` query and the new Go path side by side for a 400-vertex star, a 2000-vertex star, a polygon with a hole, a multipolygon, and a fence with a vertex placed exactly on a real pokestop, with a hostile `properties` string on every fence. All three endpoints returned identical id sets and identical quest-status counters in every case. On the 2000-vertex fence with only ~2,600 candidates the SQL path took 12.8 s and the Go path 56 ms, so your ratio reproduces on a completely different dataset.

I'll merge once the items under "Required" are addressed. The rest are suggestions; take or leave them, or split into a follow-up.

### Required

1. **New request-triggerable panic on the Go path.** `newFenceMatcher` hands every request-body polygon to `geo.CompileFence`, which reads `name` and `parent` with orb's `Properties.MustString`. In orb v0.13.0 that panics when the key is present with a non-string value, even with a default supplied. An authenticated body with `"properties":{"name":123}` now panics all three endpoints, where the old SQL path only marshalled the value into JSON. With `logging.debug` on there is no recovery middleware, so the connection drops with no response. The db path never needs `Area`, so either compile without touching properties or read them leniently.

2. **`IncDbQuery` is recorded before any row is read.** In all three streaming paths the counter sees only the `Query` open error. Scan failures, connection drops, and a mid-stream context deadline surface only through `rows.Err()` and are recorded as success. The old `Select`/`Get` calls consumed all rows first. Please pass the final error, or wrap the whole stream. Otherwise the large-fence failures this PR is about disappear from the DB error-rate dashboard.

3. **The boundary-semantics claim is wrong for MariaDB.** The description, the second commit message, and the comment on `fenceMatcher.contains` say `ST_CONTAINS` follows OGC and excludes boundary points. On 11.8.2 it returns 1 for a vertex point and for an edge midpoint, via both `ST_GeomFromText` and GeoJSON Feature input, which is why my vertex-on-a-pokestop fence produced no difference. So there is no behaviour change at all, which is better news than the PR claims, but the comment will mislead the next reader. Please correct it in all three places.

4. **clear-quests can now report a silent partial clear.** The handler runs under a 10 s context. Before, a huge fence made the SELECT itself time out and the whole call failed. Now the SELECT returns quickly and the deadline expires inside the existing per-id loop, which logs each failure and continues, then reports success. Each failed `getOrCreatePokestopRecord` also leaves a zeroed placeholder in the pokestop cache for up to an hour. The loop predates this PR, but this PR is what makes the path reachable, so please check `ctx.Err()` in the loop and return an error rather than a partial count.

### Suggested

- **Guard nil geometry.** `{"type":"Feature","geometry":null}` parses, misses the type switch, falls through to `FenceQueryArgs`, and nil-derefs on `Bound()`. Pre-existing on main, but it now lives in a new helper with no guard. A nil check in `FenceQueryArgs`/`FenceBoundArgs` (or a 400 from `NormaliseFenceFromBytes`) closes it for every caller.
- **Thread the request context** into `GetPokestopPositions` and `GetQuestStatus`. The huma handlers have one, `PokestopIdsWithinFence` already takes one, and without it an abandoned quest-status poll holds a `GeneralDb` connection and a CPU-burning goroutine until the whole result set is read and filtered.
- **Settle the bounding-box operators.** positions and quest-status use `>`/`<`, clear-quests uses `>=`/`<=`. Pre-existing drift, but `>=`/`<=` everywhere is a one-line tidy now that the predicates live next to each other.
- **The `sql.ErrNoRows` checks are dead.** `Query`/`Queryx` never return it, and `Select` into a slice and `Get` on a `COUNT(*)` don't either. Removing all four also drops the `database/sql` import.
- **Two test fixes.** Reverting one fallback branch to the old string concatenation leaves all six tests green, so the injection tests currently guard a constant rather than the query text. And both bounding-box order tests use a fence whose min lat and min lon are both 0, so swapping the min pair is undetectable; a fence like lon 10..12, lat 20..24 catches it.
- **Consider dropping the SQL fallback entirely.** It is reachable only for Point, LineString and GeometryCollection bodies. The first two are degenerate as fences (a Point matches only a coincident stop), and the third could be absorbed by flattening `orb.Collection` polygons into a MultiPolygon in `CompileFence`. Rejecting non-area geometries with a 400 at parse time would delete three fallback branches, the bind-order coupling, and the only path CI does not exercise. Happy for this to be a follow-up.

Nice work on this one.

PR #401, opened 2026-08-22
feat: cross-golbat peer lookup for pokemon stats and despawn timing

jfberry:
Adds a **cross-golbat peer lookup**: when an instance sees a wild Pokemon it cannot fully describe, it may ask a configured peer instance over gRPC and apply the answer asynchronously. Useful where instances overlap (border regions, redundant coverage) — it avoids spending a duplicate Encounter, propagates TTH-verified despawn timing, and repairs the IV loss a weather change currently inflicts on a whole S2 cell.

Configuring a peer is the switch; there is no separate flag.

```toml
[[golbat_peer]]
address = "10.0.0.2:50051"
api_secret = "shared-secret"
timeout_ms = 500
```

## Three things change for operators who configure **no** peers

Please look at these first — they are independent of the feature and land regardless.

1. **`applyVerifiedDespawn` no longer grants a phantom hour.** A sighting timestamped just past the true despawn second wrapped `+3600` and showed a just-despawned Pokemon as alive for another 59m59s. No encounter can live longer than an hour, so the wrapped value is always an impossible lifetime measured against `first_seen_timestamp`; it is clamped instead. Counted as `golbat_despawn_wrap_clamped_total`.

2. **Pokemon webhooks now also fire when `expire_timestamp_verified` changes.** Previously the trigger was new-record / species / weather / CP only, and `PokemonOldValues` carried no expiry state — so a Pokemon gaining a verified despawn emitted nothing and consumers kept the stale provisional `disappear_time`. `webhooks.md` is updated; note its dedup advice previously recommended a key that would have muted this entirely.

3. **A `despawn_sec` contradicted by a live sighting is retired.** A Pokemon seen alive past the expiry its spawnpoint implies proves that value wrong, so it is cleared and the next sighting re-derives it. This applies to every `despawn_sec`, not only peer-sourced ones, and it is the one change here that touches data Golbat learned for itself — flagging it explicitly rather than burying it. The clear is queued to a small worker because detection happens under the Pokemon lock and the clear needs the Spawnpoint lock. Counted as `golbat_despawn_retired_total`.

These three are separable and I am happy to split them into their own PRs if you would prefer to take them independently — commits `904f9ce`, `0bcc15f`, and the peer-free half of `499fb9a` respectively.

## The feature

- **One RPC**, `GetPokemon`, on the existing `Pokemon` service under the existing bearer auth. It is the gRPC counterpart of `GET /api/pokemon/id/{encounter_id}`, batched. `PokemonResult` mirrors `decoder.ApiPokemonResult`.
- **Three triggers, one question**: a Pokemon lacking IVs, one whose expiry is unverified, and one whose IVs `repopulateIv` is about to discard on a weather flip. A single lookup returns the whole record, so a Pokemon needing both stats and expiry produces one queued item.
- **The decode path never blocks.** Enqueue is a bounded channel with drop-on-full, a `util.DropReporter` and a counter, per the invariant in `CLAUDE.md`. With no peers configured the cost is two comparisons and no goroutine.
- **Answers are advisory.** A peer's `despawn_sec` is written only when the local value is NULL — local TTH always wins — and a peer's `cp`/`pvp` are never adopted, only recomputed locally from the IVs and level it supplied. Both sides verify `pokemon_id`, `form` and `weather` before an answer is used; encounter ids are reused across server-side spawn mutations, so an id alone does not identify a sighting.
- **Answering reads local cache only** and never forwards to its own peers, so lookups cannot loop.
- Locks are taken one at a time throughout, per `CLAUDE.md`.

## Deliberate omissions

- **Tappables** (`tappable_decode.go`) duplicate the second-of-hour arithmetic inline and get neither the wrap clamp nor the retire rule. Both changes scope to Pokemon on purpose; happy to extend if you would rather they were consistent.
- **Lure/map Pokemon** are not wired as a trigger. `needsExpiry` is moot there (`updateFromMap` never sets `SpawnId`) but `needsStats` is not, so this is a scope decision rather than a technical one.
- **Peer-adopted IVs are not appended to `ScanHistory`**, so a later weather flip's `locateScan` re-blanks and re-queries rather than repairing from history. Left for a dedicated change since Ditto handling and `checkScans` both read scan history.

## Testing

`go build`, `go vet`, `go test ./...` and `golangci-lint run` are clean. Around 60 new tests. `go test -race ./decoder/` is clean apart from a pre-existing race between `TestCreateStationWebhooksEmitsFutureBattle` and `statsAggregationWorker`, which reproduces on the base branch and is untouched here — worth its own issue.

One note on the generated proto: regenerating bumped the `protoc` version comment to v6.33.4. The plugin versions are pinned to what the existing files used (protoc-gen-go v1.36.11, protoc-gen-go-grpc v1.6.1), and the tree already mixes compiler versions — `raw_receiver.pb.go` was already v6.33.4 — so this aligns rather than diverges.

https://claude.ai/code/session_01QhhqwFSid9usTGHvBSE3Dx


jfberry:
will not be taken forward

Issue #402, opened 2026-08-24
Add `updated_after` to the pokemon scan filters


TurtIeSocks:
An optional `updated_after` on the pokemon scan request, so a caller can ask for only the pokemon whose `updated` is newer than a timestamp it passes in. The response shape stays the same.

## Why

ReactMap polls `/api/pokemon/v3/scan` once per connected map client. Every poll hands back the whole viewport, and nearly all of it is identical to the last one. We then diff that against what we already sent that client, so our work scales with the size of the viewport rather than with how much actually changed.

On a busy instance that is a few hundred connections re-examining one to five thousand entities every ten or twenty seconds, to find that maybe two dozen moved.

Bandwidth is not really the point here (most of us run Golbat and its consumers on the same internal network). It is the CPU on both sides, re-deriving something Golbat already knows.

## The change

`PokemonLookup` (decoder/pokemonRtree.go:34) carries no timestamp today:

```go
type PokemonLookup struct {
    PokemonId          int16
    Form               int16
    // ...
    Size               int8
    Updated            int32   // seconds, mirrors Pokemon.Updated
}
```

It gets populated in `updatePokemonLookup` (decoder/pokemonRtree.go:234), which already takes the full `*Pokemon` and runs on every change, so the value is right there:

```go
Updated: int32(pokemon.Updated.ValueOrZero()),
```

Then one comparison in `isPokemonDnfMatch` (decoder/api_pokemon_scan_v3.go), next to the existing range checks:

```go
if filter.UpdatedAfter != nil && pokemonLookup.Updated <= *filter.UpdatedAfter {
    return false
}
```

The R-tree walk in `internalGetPokemonInArea` (decoder/api_pokemon_common.go:78) does not change.

## The byte cost, which is the real question

`PokemonLookup` is 18 bytes right now, aligned to 2. Appending an `int32` pushes the alignment to 4 and the struct to 24, so it costs +6 per cached pokemon, not the +4 you would guess. Call it 6 MB per million.

Is that acceptable to you? The comment on `RaidPokemonEvolution` in `FortLookup` ("int8 in RaidLevel's pad byte keeps the struct at its pre-enrichment size") suggests these bytes get watched pretty closely. So I would rather ask than assume.

If +6 is too much, a `uint16` counting seconds from a snapshot epoch fits the existing tail padding and holds the struct at 18, at the cost of a wraparound window. Pokemon do not live long enough for that to bite. But it is more clever than a plain field, and we are not asking for it.

[...]

PR #404, opened 2026-09-09
feat: gRPC API for pokemon and fort scans (GolbatApi)

jfberry:
Adds a gRPC API service, `GolbatApi`, mirroring the HTTP scan endpoints so the API can be timed over protobuf against JSON. The gRPC path runs the same spatial index, DNF matching and per-record build as the HTTP handlers; only the serialisation differs.

Based on #399 (`perf/fortid-value-type`); retarget to `main` once that merges.

## What's here

| RPC | HTTP counterpart |
|-----|------------------|
| `ScanPokemon` | `POST /api/pokemon/v3/scan` |
| `GetPokemon` (batched, misses omitted, capped at `max_pokemon_results`) | `GET /api/pokemon/id/{pokemon_id}` |
| `ScanGyms` / `ScanPokestops` / `ScanStations` | `POST /api/{gym,pokestop,station}/scan` |
| `ScanForts` | `POST /api/fort/scan` |

- **Schema** `grpc/api.proto` (package `golbat_api`). Every message mirrors the JSON request/response struct it is named after, field for field, with proto names equal to the JSON keys. `decoder/api_grpc_parity_test.go` fails if either side gains a field the other lacks. 64-bit ids (`Pokemon.id`, `spawn_id`, `cell_id`, `encounter_ids`, the fort `cell_id`s, `bread_battle_seed`) carry `[jstype = JS_STRING]` for JS/TS clients.
- **Auth** uses the existing `api_secret` — the same value as the HTTP `X-Golbat-Secret` header — via `x-golbat-secret` metadata (`authorization` bare/Bearer also accepted). Unary and stream interceptors gate `/golbat_api.GolbatApi/*` only; raw ingest keeps its `raw_bearer` check; reflection is registered (so `grpcurl`/`ghz` work without proto files) and unauthenticated.
- **Errors** are gRPC status codes: `UNAUTHENTICATED`, `FAILED_PRECONDITION` (fort scans without `fort_in_memory`), `INVALID_ARGUMENT` (missing `min`/`max`, or too many `encounter_ids`).
- **Converters** live in `decoder/api_grpc_pokemon.go` / `api_grpc_fort.go`. The pokemon result loop is now a shared visitor (`forEachLivePokemonResult`) used by both the JSON collector and the gRPC one, so the peek/expiry/build logic exists once. Stored JSON blobs (defenders, quest rewards, showcase data, stationed pokemon) pass through as `*_json` text exactly as the HTTP API emits them.
- **Docs**: `api.md` gRPC section (semantic differences from JSON, message-size guidance, benchmarking caveats), `CLAUDE.md`, `config.toml.example`. Spec and plan under `docs/superpowers/`.

## Breaking change

The stub `Pokemon` gRPC service (`Search`, `SearchV3`, five fields per result) is **removed**; `GolbatApi` replaces it. `grpc/raw_receiver_grpc.pb.go` was regenerated with the pinned `protoc-gen-go-grpc` 1.6.1 (it was on 1.5.1): version comment, a generated-comment typo, and `status.Errorf` → `status.Error` with no verbs. A fresh `update_grpc.sh` run now produces no diff.

## Things a client author should know

- `IntRange`: unset `max` means no upper bound (32767), unlike JSON where an omitted `max` is 0. When benchmarking against HTTP, send both bounds or neither.
- Pokemon `filters`: an empty list matches nothing (JSON parity); one clause with no conditions matches everything. Fort `filters`: empty matches every fort of the type (also JSON parity).
- Repeated list filters: empty means no constraint (proto3 cannot express the JSON "explicitly empty list matches nothing" case).
- Fort scans can exceed the grpc-go client default 4 MB receive limit at `max_fort_results = 9000`; raise `MaxCallRecvMsgSize` (documented in `api.md`).

## Follow-ups (not in this PR)

- gRPC Prometheus metrics are created and initialised but never registered with a registry (pre-existing).
- `int32sTo` / contest `min_level` truncate rather than clamp out-of-range filter values (values are small enums today).

## Testing

`go build`, `go vet`, `go test ./...` clean; `go test -race` clean on `.`, `./grpc/` and the gRPC decoder tests (the full `-race ./decoder/` has a pre-existing unrelated station-stats flake). New tests: golden converter tests for every message, request-conversion tests, the reflection parity test, live scan tests through the real tree/cache/endpoint path, interceptor unit tests, and bufconn end-to-end tests of the production server construction (auth matrix, fort gate, validation, GetPokemon cap, Prometheus-chain ordering). `golangci-lint` 2.12.2 cannot load Go 1.27 modules here, so lint did not run.

https://claude.ai/code/session_013WpTQXBGa8jJnbxVJo2EqY


jfberry:
Added `battle_available` to the fort DNF (2d528b3): mirrors the station's `is_battle_available` flag only. `station_active` now means currently active (`is_inactive = 0 AND start_time < now AND end_time > now`; it previously checked `end_time` alone), so "battle available now" is `battle_available:true AND station_active:true`. The three lookup fields sit in `FortLookup`'s tail padding — `Sizeof` stays 184, pinned by a test. Proto: `optional bool battle_available = 22`; `/api/station/available` advertises `battle_available_filter`.

jfberry:
Combined fort scan now takes a per-type `limit` on each group (`FortTypeScanGroup.limit = 2`, JSON `limit`) and reports `gyms_stats` / `pokestops_stats` / `stations_stats` (`examined`, `limit_reached`) per type (b24956a). A type stops accepting matches at its own limit while the walk keeps filling the others; the walk ends when every requested type is full or the top-level `limit` (still an overall cap) is hit. Top-level `examined` / `skipped` / `total` unchanged; top-level `limit_reached` now means any type or the overall cap. Excluded types report zero examined and `limit_reached=false`. No `FortLookup` change.

jfberry:
Capability flags now live on `/api/status` under a new `filters` block (4533966):

```json
{ "features": {...}, "limits": {...}, "filters": { "showcase_focus": true, "battle_available": true } }
```

A consumer sends a filter only when its key is present and true; an older Golbat omits the block. `showcase_focus_filter` is removed from `/api/pokestop/available` (and the nested copy in `/api/fort/available`), and the short-lived `battle_available_filter` on `/api/station/available` is gone too — availability responses list active filter options, not capabilities. Diadem should read `status.filters.showcase_focus` instead of the availability flag.

jfberry:
Correction to the previous note: `showcase_focus_filter` **stays** on `/api/pokestop/available` (and nested in `/api/fort/available`), marked deprecated (58-char doc). ReactMap's `Pokestop.getAvailable` throws "Golbat lacks the required showcase_focus_filter capability" when the key is missing or false and deliberately does not fall back to SQL, and ReactMap has no `/api/status` plumbing. `status.filters.showcase_focus` is the forward path; `battle_available` is advertised only there. Nothing else changes.

jfberry:
`updated_after` landed (d778452), closing #402: on the pokemon v3 scan and every fort scan (HTTP and gRPC), strict `updated > updated_after`, applied at response build on the locked record so no lookup bytes. Counters and `limit_reached` describe the pre-gate scan (documented); advertised as `status.filters.updated_after`.

